PR #123, opened 2026-02-07
Expression, gradient and tooling improvements

ashesfall:
This PR extends the Almost Realism tooling and core expression system to better reproduce and diagnose CI-only gradient failures (especially around sparse-Jacobian masked sums), while also improving developer workflows around test execution and memory introspection.

Changes:

Adds guardrails and knobs around masked-sum reordering and gather-collapse analysis to prevent pathological compile-time work in deep/shared gradient DAGs.
Refactors Expression by extracting mixins and adds optional per-evaluation value memoization (flag-gated) for shared-DAG evaluation.
Improves MCP tooling: CI test-group reproduction in the test runner, artifact-age preflight reporting, and a new “memory namespaces” summary API/tooling.


flowtree/runtime/src/main/java/io/flowtree/Agent.java
```diff
@@ -0,0 +1,46 @@
...
+ * for a single-job agent participating in a peer network.</p>
+ */
+public class Agent {
+    /**
+     * Launches the agent server with default properties.
+     *
+     * @param args command-line arguments (unused)
+     * @throws IOException if the server fails to start
+     */
+    public static void main(String[] args) throws IOException {
```
ashesfall: We don't need this class, it should not be reintroduced after removal on master.

base/code/src/main/java/io/almostrealism/expression/Conditional.java
```diff
@@ -169,6 +169,60 @@ public static Expression of(Expression<Boolean> condition, Expression<?> positiv
...
+		return new DirectConditional(condition, positive, negative);
+	}
+
+	/**
+	 * A Conditional that preserves itself through simplification without being
+	 * converted to a {@link Mask}. Overrides {@link #recreate(List)} to bypass
+	 * the {@link Conditional#create(Expression, Expression, Expression)} factory
+	 * which converts conditionals with zero negative branches to Mask expressions.
+	 */
+	private static class DirectConditional extends Conditional<Double> {
```
ashesfall: This should not exist - we should either find a way to intelligently determine when it makes sense to replace Conditionals with Masks, or we should accept that the conversion to Mask is counter-productive and turn it off.

base/code/src/main/java/io/almostrealism/profile/OperationMetadata.java
```diff
@@ -289,6 +289,33 @@ public OperationMetadata appendShortDescription(String desc) {
...
 
+	/**
+	 * Mutates this metadata's shortDescription to indicate provenance.
+	 * The new shortDescription follows the pattern: "parent ==> child"
+	 * where child is the current shortDescription (or displayName if no
+	 * shortDescription is set).
+	 *
+	 * @param parent the parent context name to prepend
+	 */
+	public void applyProvenance(String parent) {
```
ashesfall: This should not be allowed.

PR #175, opened 2026-03-23
AudioScene redesign

ashesfall:


studio/music/src/main/java/org/almostrealism/music/pattern/PatternLayerManager.java
```diff
@@ -126,6 +126,45 @@ public class PatternLayerManager implements PatternFeatures, HeredityFeatures {
...
+	 */
+	private static final int BATCHED_FILTER_ORDER = 40;
+
+	/**
+	 * Audio sample rate used when constructing a per-pattern
+	 * {@link BatchedPatternLayerRenderer}. Matches the production
+	 * {@code OutputLine.sampleRate} value used by every existing pattern
+	 * acoustic-equivalence test.
+	 */
+	private static final int BATCHED_SAMPLE_RATE = 44100;
```
ashesfall: In the application where we actually use all this code, OutputLine.sampleRate is set programmatically at startup based on the users configured sample rate. You have only two options when performing calculations that assume a sample rate: use the *actual value* of OutputLine.sampleRate, or accept the sample rate as a parameter which can be provided by the caller. ANY constant that controls the sample rate is invalid by definition (44.1 isn't even the most common sample rate that users actually use, its basically just the lowest option we provide in the selector and so we default to it for tests - committing to it like this breaks EVERYTHING. If you were even going to assume a sample rate, which you never should, it would be far higher than 44.1)

ashesfall: Acknowledged — committing to a sample-rate constant is the wrong shape; the value has to come from `OutputLine.sampleRate` (or be passed in by the caller). The hardcoded 44100 was a mistake.

Fixed at `6e2e45ffb3`: every occurrence of the literal `44100` and the duplicate filter-order constant `40` now reads from the canonical source.

Specifically:
- `studio/music/.../PatternLayerManager.java` — `BATCHED_SAMPLE_RATE = 44100` → `OutputLine.sampleRate`; `BATCHED_FILTER_ORDER = 40` → `MultiOrderFilterEnvelopeProcessor.filterOrder`.
- `engine/audio/.../BatchedPatternRendererTest.java` and `engine/audio/.../PatternRenderingFloorBenchmark.java` — both `SAMPLE_RATE = 44100` references → `OutputLine.sampleRate`.

On the filter-order constant: `EfxManager.filterOrder` would be the more natural canonical source for production, but `EfxManager` lives in `ar-compose` (studio/compose) which `ar-music` (studio/music) cannot depend on without inverting the layer order. `MultiOrderFilterEnvelopeProcessor.filterOrder` is the engine/audio-resident constant that's actually reachable from this module's classpath, so I used it as the canonical reference for both the production code in `ar-music` and the engine/audio tests. If/when the filter-order constant should live on `EfxManager` instead, the right move is to relocate it into a lower-layer module — happy to do that as a follow-on if you'd prefer.

Two adjacent constants in the same area were flagged but deliberately left unchanged: `BATCHED_SOURCE_LENGTH = 2048` and `BATCHED_TARGET_LENGTH = 1024` in `PatternLayerManager`. These came in as benchmark-derived fixtures and have no current canonical source, but baking a fixed source/target length into the per-tick pattern renderer is the same anti-pattern at a different layer — pattern-element rendering probably shouldn't have fixed buffer lengths either. Calling that out as a separate cleanup rather than rolling it into this fix.

Post-fix Mac/Metal benchmark numbers (new "Addition 8" in `docs/plans/audio-scene-redesign/PATTERN_RENDERING_FLOOR.md`, captured on `Mac OS X / aarch64 / java 24.0.1`, `OutputLine.sampleRate = 44100`) match the prior measurements — E2/E3 at 64 notes/m run 25–27% faster, B1 13–17% faster, never slower. The fix is a no-op at the default sample rate (which is what `BufferedAudio.sampleRate` resolves to in tests), exactly as expected.

studio/music/src/main/java/org/almostrealism/music/pattern/PatternFeatures.java
```diff
@@ -159,13 +248,72 @@ default void render(AudioSceneContext sceneContext, NoteAudioContext audioContex
...
+	 *
+	 * @param noteStart           the frame offset of the note that failed
+	 * @param consecutiveFailures running count of failures since the last success
+	 * @param cause               the exception thrown by the evaluation
+	 */
+	private void handleNoteEvaluationFailure(int noteStart, int consecutiveFailures,
+											 Exception cause) {
+		warn("Note evaluation failed at frame " + noteStart + ": " + cause.getMessage());
+
+		if (isExhaustionFailure(cause)) {
```
ashesfall, line 267: We should adjust the underlying computation engine to throw a specific type of exception in this case, rather than using string matching.

studio/music/src/main/java/org/almostrealism/music/pattern/PatternFeatures.java
```diff
@@ -159,13 +248,72 @@ default void render(AudioSceneContext sceneContext, NoteAudioContext audioContex
...
+	}
+
+	/**
+	 * Walks the cause chain of the given throwable and returns {@code true} if
+	 * any link reports a message matching {@link #NOTE_EVAL_EXHAUSTION}.
+	 *
+	 * @param t the throwable to inspect
+	 * @return {@code true} when the chain reports operator-pool exhaustion
+	 */
+	private static boolean isExhaustionFailure(Throwable t) {
```
ashesfall, line 293: We should adjust the underlying computation engine to throw a specific type of exception in this case, rather than using string matching.

base/code/src/main/java/io/almostrealism/expression/Ceiling.java
```diff
@@ -145,6 +145,38 @@ public Expression<Double> recreate(List<Expression<?>> children) {
...
+	 * @return a constant, the operand unchanged, or a new {@link Ceiling}
+	 */
+	public static <T> Expression<T> of(Expression in) {
+		OptionalDouble d = in.doubleValue();
+
+		if (d.isPresent()) {
+			return (Expression<T>) new DoubleConstant(Math.ceil(d.getAsDouble()));
+		}
+
+		if (in.getType() == Integer.class || in.getType() == Long.class) {
```
ashesfall: This recommended way to check this is using !in.isFP() (this will also cover Boolean, which you can filter out separately if needed, but Ceiling of a boolean should also just be the boolean value (assuming this could ever even happen) so it should be fine.

base/code/src/main/java/io/almostrealism/expression/Floor.java
```diff
@@ -84,22 +84,38 @@ public Expression<Double> recreate(List<Expression<?>> children) {
...
-	public static Expression of(Expression in) {
+	public static <T> Expression<T> of(Expression in) {
 		OptionalDouble d = in.doubleValue();
 
 		if (d.isPresent()) {
-			return new DoubleConstant(Math.floor(d.getAsDouble()));
+			return (Expression<T>) new DoubleConstant(Math.floor(d.getAsDouble()));
+		}
+
+		if (in.getType() == Integer.class || in.getType() == Long.class) {
```
ashesfall: This recommended way to check this is using !in.isFP() (this will also cover Boolean, which you can filter out separately if needed, but Ceiling of a boolean should also just be the boolean value (assuming this could ever even happen) so it should be fine.

ashesfall: Done. Replaced the explicit `Integer.class || Long.class` check with `!in.isFP()` in `Ceiling.of()`. This is broader — it also covers Boolean and any future non-FP type without having to enumerate them. `ExpressionSimplificationTests` (67 tests) still passes.

base/code/src/main/java/io/almostrealism/expression/Ceiling.java
ashesfall: Done. Same change applied to `Floor.of()` — the explicit type enumeration is replaced with `!in.isFP()`. `ExpressionSimplificationTests` (67 tests) still passes.

studio/music/src/main/java/org/almostrealism/music/pattern/PatternFeatures.java
ashesfall: Done. Created `OperatorPoolExhaustedException extends HardwareException` in `org.almostrealism.hardware`, updated `AcceleratedOperation.load()` to throw it instead of the generic `HardwareException`, and replaced the regex string-match in `isExhaustionFailure` with an `instanceof` walk of the cause chain. The test was also updated to throw and catch the specific type. `PatternFeaturesFastFailTest` (both exhaustion and consecutive-failure cases) passes.

studio/music/src/main/java/org/almostrealism/music/pattern/PatternFeatures.java
ashesfall: Agreed — the regex was fragile. Replaced with a typed `OperatorPoolExhaustedException` that `AcceleratedOperation` now throws directly, and `isExhaustionFailure` checks `instanceof OperatorPoolExhaustedException` while walking the cause chain. No string matching anywhere in the path. `PatternFeaturesFastFailTest` confirms the fast-fail behavior is intact.

PR #193, opened 2026-04-15
Audio DSP using PDSL

ashesfall:


engine/ml/src/main/java/org/almostrealism/ml/dsl/PdslInterpreter.java
```diff
@@ -963,6 +1066,306 @@ private Object callRange(List<Object> args) {
...
 	}
 
+	/**
+	 * Returns a block factory that wraps the given FIR coefficients in a {@link MultiOrderFilter}.
+	 *
+	 * @param name         primitive name used as the layer label
+	 * @param coefficients pre-computed FIR coefficient producer
+	 * @return a factory that creates a FIR filter block for any input shape
+	 */
+	private Function<TraversalPolicy, Block> firFilterBlock(String name, CollectionProducer coefficients) {
```
ashesfall: All of these Audio DSP utility methods should be located elsewhere, the PdslInterpreter class should not be cluttered by this kind of thing.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/PdslApproachDValidationTest.java
```diff
@@ -0,0 +1,231 @@
...
+ *       {@link Temporal#tick()} invocations when the block is wrapped as a Temporal.</li>
+ *   <li><strong>Block→Temporal adapter forward flow</strong> — a minimal Temporal
+ *       wrapping {@link CompiledModel#forward} channels input through to output.</li>
+ * </ol>
+ *
+ * @see org.almostrealism.ml.dsl.PdslLoader
+ * @see org.almostrealism.audio.CellList
+ * @see org.almostrealism.time.Temporal
+ */
+public class PdslApproachDValidationTest extends TestSuiteBase implements FirFilterTestFeatures {
```
ashesfall: The title of this test makes no sense, "approach D" is just an artifact from the planning document not a long term concept of the platform.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/PdslApproachDValidationTest.java
```diff
@@ -0,0 +1,231 @@
...
+	public void testRequirementTicksOncePerBufferNotPerSample() {
+		AtomicInteger tickCallCount = new AtomicInteger(0);
+
+		Temporal requirement = () -> {
+			tickCallCount.incrementAndGet();
+			return () -> () -> {};
+		};
+
+		CellList list = new CellList();
+		list.setTickLoopCount(BUFFER_SIZE);
```
ashesfall: This method no longer exists, and is basically no longer necessary either.

engine/ml/src/main/java/org/almostrealism/ml/dsl/PdslInterpreter.java
ashesfall: Done. Extracted every audio DSP `call*` method (`callFir`, `callScale`, `callIdentity`, `callLowpass`, `callHighpass`, `callBiquad`, `callDelay`, `callLfo`, plus the `firFilterBlock` factory) into a new `AudioDspInterpreterFeatures` mixin interface alongside `MultiChannelDspFeatures`. The `tryCallBuiltin` ladder was replaced with a registration/lookup table — `PdslInterpreter` holds a `Map<String, Function<List<Object>, Object>>` that the features interface populates via `registerAudioPrimitives`, so adding a new audio primitive no longer requires editing the interpreter. The multi-channel `callRoute`/`callSumChannels`/`callFanOut` dispatchers also moved to the new interface (they take `channels` and `signal_size` as plain ints from the environment), while `interpretForEachChannel` stayed in `PdslInterpreter` since it is a language-level `Statement` handler — a sibling of `interpretBranch`/`interpretAccum` that recurses through `interpretBody`. `PdslInterpreter.java` no longer mentions `callBiquad`, `callDelay`, etc., and dropped from 1600 to 1334 lines.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/PdslApproachDValidationTest.java
ashesfall: Renamed to `PdslLayerCellListIntegrationTest` (the tests exercise PDSL-compiled blocks integrating into the CellList path via `Temporal.tick`). All "Approach D" terminology removed from class name, method names, javadoc, and the one remaining reference in `docs/plans/PDSL_AUDIO_DSP.md`. The remaining tests run green at depth 2.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/PdslApproachDValidationTest.java
ashesfall: Removed. The reference at line 97 was `CellList.setTickLoopCount`, which no longer exists; the test (`testRequirementTicksOncePerBufferNotPerSample`) is deleted entirely per your "no longer necessary either" note. The other two tests in the class — state persistence across `Temporal.tick` and the Block→Temporal adapter forward flow — exercise the integration surface the renamed class is named for, and remain in place.

PR #194, opened 2026-04-16
Corrected code policy violations

ashesfall:


engine/ml/src/main/java/org/almostrealism/ml/dsl/PdslInterpreter.java
```diff
@@ -1153,6 +1154,24 @@ private Producer<PackedCollection> toProducer(Object value) {
...
 
+	/**
+	 * Converts a value to a {@link CollectionProducer} for use as a RoPE frequency tensor.
+	 * Accepts either a {@link PackedCollection} (wrapped via {@code cp()}) or an existing
+	 * {@link CollectionProducer}.
+	 *
+	 * @param value the argument value from the PDSL args map
+	 * @return a CollectionProducer wrapping the frequency tensor
+	 */
+	private CollectionProducer toFreqCis(Object value) {
```
ashesfall: This is absurd, there is nothing "frequency cis" specific about this method it is just a general purpose wrap operation - correct the name.

engine/utils/src/main/java/org/almostrealism/util/ProducerPatternDetector.java
```diff
@@ -55,6 +57,17 @@ public class ProducerPatternDetector extends PolicyViolationDetector {
 			"/studio/"
 	);
 
+	/**
+	 * Methods where {@code .evaluate()} calls are legitimate despite being in
+	 * a computation source tree. These are autoregressive step loops or other
+	 * boundaries where the compiled model is invoked token-by-token.
+	 */
+	private static final List<String> EVALUATE_ALLOWED_METHODS = List.of(
```
ashesfall: This should not exist: all method exceptions should be done at the level of a specific class or file.

PR #201, opened 2026-04-20
Broken MIDI Model Tests

ashesfall:


base/relation/src/main/java/io/almostrealism/compute/Process.java
```diff
@@ -211,6 +215,62 @@ default long getOutputSize() {
 		return 0;
 	}
 
+	/**
+	 * Returns the factor by which inlining this process's emitted expression
```
ashesfall: Please correct the language of this comment so that it does not refer specifically to Expressions - the entire reason we use the term "expansion width" is to generalize to all Process implementations, not just those which use Expressions.

compute/algebra/src/main/java/org/almostrealism/collect/computations/ReshapeProducer.java
```diff
@@ -440,6 +442,20 @@ public ReshapeProducer generate(List<Process<?, ?>> children) {
 
 	@Override
 	public ParallelProcess<Process<?, ?>, Evaluable<? extends PackedCollection>> optimize(ProcessContext ctx) {
+		ProcessOptimizationStrategy strategy = ctx.getOptimizationStrategy();
+
+		if (strategy != null && producer instanceof Process) {
```
ashesfall: The only way this could possibly be necessary is if optimize(ctx, producer) below fails to respect the optimization strategy, which is either definitely false (I think) or (in the very unlikely event it is true) would need to be corrected rather than having this weird one-off customization.

ashesfall: You're correct — `ParallelProcess.super.optimize(ctx)` (via `CollectionProducerParallelProcess`) already invokes the strategy on `this`, making the explicit `strategy.optimize(context, this, ...)` block redundant. I've simplified the override to delegate to `CollectionProducerParallelProcess.super.optimize(ctx)`.

While tracing this I also found that `ReshapeProducer` was missing a `getExpansionWidth()` override — its wrapped `producer` would report the correct expansion width (e.g. `CollectionProductComputation` returns 2 for a binary multiply) but the `ReshapeProducer` was returning the default 1, hiding the expansion from the optimization context. I've added a delegation: `return producer instanceof Process ? ((Process<?, ?>) producer).getExpansionWidth() : 1;`, which matches the plan in `docs/plans/EXPANSION_WIDTH_FAITHFUL_DEFINITION.md`. `ExpansionWidthTests` and `ReshapeProducerStrategyTests` both pass (9/9).

base/relation/src/main/java/io/almostrealism/compute/Process.java
ashesfall: Done — the `getExpansionWidth()` Javadoc in `Process.java` has been rewritten to use general process-level language. It no longer mentions Expressions; it talks about "inlining this process into a consumer", "per-use-site contribution", and "output size growth at compile time", which applies equally to all `Process` implementations.

PR #211, opened 2026-04-26
Fixing issues required for AudioScene rendering

ashesfall:


studio/compose/scripts/setup-reverb-experiment.sh
```diff
@@ -0,0 +1,23 @@
...
+# Removes the population JSON and per-run output directories so that the
+# next optimizer launch starts from a fresh genome, then materialises the
+# two destination directories (`health-reverb-off` and `health-reverb-on`)
+# that hold WAVs from each leg of the experiment.
+#
+# Usage:
+#   scripts/setup-reverb-experiment.sh
+set -euo pipefail
+
+WORKING_DIR="/Users/Shared/Music"
```
ashesfall: This script can be removed now that we have completed our work.

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManager.java
```diff
@@ -958,8 +978,13 @@ public Configuration(int scale) {
 			overallFilterUpOffsetMin = offset + 15.0;
 			overallFilterUpOffsetMax = offset + 45.0;
 
-			minTransmission = 0.3;
-			maxTransmission = 0.6;
+			// Tightened from [0.3, 0.6]. With 4 delay layers and the mself
```
ashesfall: Probably this class can just be restored to its original form, as we have now fixed the issues causing us trouble in other ways.

studio/compose/src/main/java/org/almostrealism/studio/optimize/AudioSceneOptimizer.java
```diff
@@ -350,7 +350,16 @@ public static void run(OperationProfileNode profile) {
 		// auto-volume step required look-ahead which the realtime pipeline
 		// can't provide), so scale the summed master bus to give it
 		// headroom and stop the int16 WAV writer from clipping.
-		MixdownManager.masterBusGain = 0.2;
+		// Master gain at 0.5 with the DelayNetwork normalisation fix in
+		// place. This gives the master a strong listening level (typically
+		// -10 to -14 dBFS RMS) while leaving the limiter enough room to
+		// catch transient peaks without constantly clipping.
+		MixdownManager.masterBusGain = 0.5;
```
ashesfall: We should not be customizing these values here. If 0.5 seems good, we should make it the default.

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManager.java
```diff
@@ -173,6 +173,15 @@ public class MixdownManager implements Setup, Destroyable, CellFeatures, Optimiz
...
 	public static double masterBusGain = 1.0;
 
+	/**
+	 * Gain applied to the summed EFX bus before it is mixed into the main
+	 * signal. The EFX delay grid sums the forward path across (channels x
+	 * delay layers) which can easily push the bus level past unity even
+	 * before any feedback, so this defaults below 1.0 to keep the bus from
+	 * dominating the master mix.
+	 */
+	public static double efxBusGain = 1.0;
```
ashesfall: We don't need this field, we are only ever setting it to 1.0.

studio/compose/src/main/java/org/almostrealism/studio/optimize/AudioSceneOptimizer.java
```diff
@@ -350,7 +350,16 @@ public static void run(OperationProfileNode profile) {
...
+		// efxBusGain disabled (=1.0) for this test -- bounded transmission
+		// and wetOut chromosome ranges should be enough to keep the EFX bus
+		// stable on their own. If the EFX channel is still divergent, the
+		// chromosome bound mechanism isn't reaching the kernel.
+		MixdownManager.efxBusGain = 1.0;
```
ashesfall: This is the default, and we should not be assigning it.

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManager.java
```diff
@@ -693,6 +703,16 @@ public CellList createEfx(CellList main, CellList efx, CellList reverb,
 			}
 		}
 
+		if (efxBusGain != 1.0) {
```
ashesfall: This is never used.

PR #212, opened 2026-04-26
MixdownManager implementation using PDSL

ashesfall:


engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
```diff
@@ -175,37 +238,78 @@ default Object callLowpass(List<Object> args) {
...
+	 * {@code lowPassCoefficients} / {@code highPassCoefficients} helpers. Both
+	 * helpers already accept a {@link Producer}, so a literal is wrapped via
+	 * {@code c(value)} and a producer is forwarded directly without an extra
+	 * {@code subset} (the helpers handle their own shape arithmetic).
+	 *
+	 * @param arg       the cutoff argument from the PDSL argument list
+	 * @param primitive primitive name used in the error message
+	 * @return the cutoff as a {@link Producer} of {@link PackedCollection}
+	 */
+	default Producer<PackedCollection> cutoffProducer(Object arg, String primitive) {
```
ashesfall: Having a dedicated method for this is stupid, there is nothing special about the cutoff frequency - we can't just introduce dedicated methods for every imaginable argument.

engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
```diff
@@ -70,6 +70,36 @@ static double toDouble(Object value) {
...
+	 * @param arg the scalar argument from the PDSL argument list
+	 * @return a shape-{@code [1]} {@link CollectionProducer} suitable for use in
+	 *         broadcast multiplication, addition, and subtraction inside a kernel
+	 * @throws PdslParseException if {@code arg} is neither a {@link Number} nor a {@link Producer}
+	 */
+	default CollectionProducer scalarProducer(Object arg) {
+		if (arg instanceof Number) {
+			return c(toDouble(arg));
+		}
+		if (arg instanceof Producer) {
```
ashesfall, line 116: This should not hide the fact the producer is the wrong shape by extracting the first value out of it: if a producer that is not the right shape is being used, an exception should be thrown. CollectionFeatures::shape() can be used to identify the shape of any Producer.

engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
```diff
@@ -70,6 +70,36 @@ static double toDouble(Object value) {
...
+	 * {@code delay}, {@code lfo}, {@code lowpass}, {@code highpass}). It centralizes
+	 * the literal-vs-producer branch so each primitive only has to call it for
+	 * each of its scalar arguments.</p>
+	 *
+	 * @param arg the scalar argument from the PDSL argument list
+	 * @return a shape-{@code [1]} {@link CollectionProducer} suitable for use in
+	 *         broadcast multiplication, addition, and subtraction inside a kernel
+	 * @throws PdslParseException if {@code arg} is neither a {@link Number} nor a {@link Producer}
+	 */
+	default CollectionProducer scalarProducer(Object arg) {
```
ashesfall: How could this method be necessary when we have a fully general solution in PdslInterpreter.bindProducerParameter ?

engine/ml/src/main/java/org/almostrealism/ml/dsl/PdslInterpreter.java
```diff
@@ -174,13 +174,71 @@ public Block buildLayer(String name, TraversalPolicy inputShape,
...
+
+		if (value instanceof Number) {
+			if (declaredShape.getTotalSize() != 1) {
+				throw new PdslParseException(
+						"Parameter '" + param.getName() + "' has declared shape "
+								+ declaredShape + " but a Number literal can only be supplied for shape [1]");
+			}
+			PackedCollection wrapped = new PackedCollection(declaredShape);
+			wrapped.setMem(0, ((Number) value).doubleValue());
+			return FEATURES.cp(wrapped);
```
ashesfall: Why would you use cp(collection) (reference a collection) for this when you could use c(value) (an explicit constant value). The only reason to use p() or cp() is if the argument value needs to change over subsequent executions, and you know in this case that it can't.

engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
```diff
@@ -70,6 +71,66 @@ static double toDouble(Object value) {
...
+			PackedCollection coll = (PackedCollection) arg;
+			if (coll.getShape().getTotalSize() != expectedShape.getTotalSize()) {
+				throw new PdslParseException(contextName + " expects shape "
+						+ expectedShape + " but PackedCollection has shape " + coll.getShape());
+			}
+			return cp(coll);
+		}
+		if (arg instanceof Producer) {
+			Producer<PackedCollection> producer = (Producer<PackedCollection>) arg;
+			if (producer instanceof Shape) {
```
ashesfall: This is not necessary, you can just use shape(producer) to get the shape.

engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
```diff
@@ -70,6 +71,66 @@ static double toDouble(Object value) {
...
+				TraversalPolicy actual = ((Shape<?>) producer).getShape();
+				if (actual.getTotalSize() != expectedShape.getTotalSize()) {
+					throw new PdslParseException(contextName + " expects shape "
+							+ expectedShape + " but Producer has shape " + actual);
+				}
+			}
+			if (producer instanceof CollectionProducer) {
```
ashesfall: This is also not necessary, c(producer) already includes this logic.

PR #216, opened 2026-04-29
MixdownManager as PDSL

ashesfall:


engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
```diff
@@ -164,16 +172,33 @@ default Function<TraversalPolicy, Block> firFilterBlock(String name,
...
+	 *             as a {@link PackedCollection}, {@link CollectionProducer}, or
+	 *             {@link Producer}{@code <PackedCollection>}
 	 * @return a factory that creates a FIR filter block for any input shape
 	 */
 	default Object callFir(List<Object> args) {
-		if (args.size() == 1 && args.get(0) instanceof PackedCollection) {
-			PackedCollection coefficients = (PackedCollection) args.get(0);
-			return firFilterBlock("fir", cp(coefficients));
+		if (args.size() == 1) {
+			Object arg = args.get(0);
```
ashesfall: This handling of collection, versus collection producer, versus general producer should not be rewritten over and over - it should be a shared utility.

engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
```diff
@@ -480,12 +505,37 @@ default Object callLfo(List<Object> args) {
...
 	 */
 	default Block callRoute(List<Object> args, int inputChannels, int signalSize) {
-		if (args.size() != 1 || !(args.get(0) instanceof PackedCollection)) {
+		if (args.size() != 1) {
 			throw new PdslParseException(
-					"route() expects 1 weight argument (transmission matrix), got " + args.size());
+					"route() expects 1 transmission matrix argument, got " + args.size());
+		}
+		Object arg = args.get(0);
+		if (arg instanceof PackedCollection) {
```
ashesfall: It is completely absurd to do this twice, once for the collection and once for the producer. Just cover the CollectionProducer case, and use a shared utility to ensure that all collections are properly wrapped as producers. This is a common feature of a bunch of ugly code - treating collections and producers with separate paths when all you need is one path and a single line check to ensure that all collections are wrapped.

engine/ml/src/main/java/org/almostrealism/ml/dsl/MultiChannelDspFeatures.java
```diff
@@ -114,6 +114,27 @@ default Block perChannelBlock(List<Block> channelBlocks, int channels, int signa
 	 */
 	default Block routeBlock(PackedCollection matrix, int inputChannels,
```
ashesfall: Don't even permit this, allowing a collection that is not wrapped as a Producer just invites bad behavior.

engine/ml/src/main/resources/pdsl/audio/mixdown_manager.pdsl
```diff
@@ -320,4 +362,66 @@ layer mixdown_master(channels: int, signal_size: int,
...
+//
+// Requires mixdown_reverb_state populated with:
+//   reverb_buffers   PackedCollection of size channels * signal_size (zero-init)
+//   reverb_heads     PackedCollection of size channels (zero-init)
+// ---------------------------------------------------------------------------
+layer mixdown_reverb_bus(channels: int, signal_size: int,
+                         delay_samples: producer([channels]),
+                         feedback_matrix: producer([channels, channels]))
+                      -> [1, signal_size] {
+    fan_out(channels)
```
ashesfall: We have a Block operation called repeat. How is fan_out any different than repeat? Why are we inventing something new here?

engine/ml/src/main/java/org/almostrealism/ml/dsl/AudioDspInterpreterFeatures.java
ashesfall: This class, like all audio specific capabilities, should live in an audio-specific module in the project. The reason it was put here is, I assume, because PdslInterpreter *calls it directly*. But this is a huge mistake: PdslInterpreter should not have to know about every single possible operation, it should be possible for capabilities like those provided here to be *registered* with PdslInterpreter so that the core PDSL code can live in a module like ar-graph (or ar-ml, if we can't push it that far down the dependency tree) while additional capabilities are added in higher-level modules via registration.

engine/ml/src/main/java/org/almostrealism/ml/dsl/PdslPrimitive.java
```diff
@@ -0,0 +1,59 @@
...
+public interface PdslPrimitive {
+
+	/**
+	 * Build the result of a single primitive call.
+	 *
+	 * @param args the evaluated argument list from the PDSL source
+	 * @param ctx  the per-call context (channels, signal size, normalisation)
+	 * @return the result the interpreter appends to the surrounding block
+	 */
+	Object dispatch(List<Object> args, PdslPrimitiveContext ctx);
```
ashesfall: A primitive should only produce one type of thing, so the interface should use a generic argument to decide what that is and this dispatch method should return the matching type. eg PdslPrimitive<T> should have a dispatch method returning T.

studio/compose/src/main/java/org/almostrealism/studio/dsl/audio/AudioDspPrimitives.java
```diff
@@ -0,0 +1,368 @@
...
+	private static final AudioDspPrimitives INSTANCE = new AudioDspPrimitives();
+
+	/**
+	 * Register every audio primitive and the multi-channel dispatcher with
+	 * {@code interpreter}. Idempotent: re-registering replaces any previous
+	 * binding for the same primitive name.
+	 *
+	 * @param interpreter the freshly-built interpreter to configure
+	 */
+	public static void registerWith(PdslInterpreter interpreter) {
```
ashesfall, line 89: There are a lot of primitives here that are not audio specific, and that I believe are redundant: scale (just multiply by a constant?), sum_channels (isn't this just accum?), fan_out (this should be named "repeat" to be consistent with the CollectionProducer method naming, and it is definitely not audio-specific). Some of the others are suspect as well.

studio/compose/src/main/java/org/almostrealism/studio/dsl/audio/AudioDspPrimitives.java
```diff
@@ -0,0 +1,368 @@
...
+			throw new PdslParseException("scale() expects 1 argument (factor), got " + args.size());
+		}
+		CollectionProducer factor = ctx.toProducer(args.get(0), shape(1), "scale() factor");
+		return (Function<TraversalPolicy, Block>)
+				(shape -> layer("scale", shape, shape,
+						input -> multiply(c(input).each(), factor)));
+	}
+
+	/** {@code identity()} — pass-through block. */
+	private Object dispatchIdentity(List<Object> args, PdslPrimitiveContext ctx) {
```
ashesfall: It's concerning that we would ever require this, but if we do it should: (1) not be here (it isn't audio specific, put it in a more general place) and (2) should support both backward and forward (why null out backward for no reason?)

PR #220, opened 2026-05-07
Multi-Layer Audio Format

ashesfall:


studio/compose/src/main/proto/audio.proto
```diff
@@ -67,4 +70,190 @@ message SynthesizerModelData {
   repeated double frequency_ratios = 2;
   double level_sample_rate = 3;
   CollectionData level_data = 4;
-}
\ No newline at end of file
+}
+
+// =========================================================================
+// Multi-layer audio extensions (Common 0.74)
```
ashesfall: Do not put version numbers in files like this

studio/compose/src/main/proto/audio.proto
```diff
@@ -67,4 +70,190 @@ message SynthesizerModelData {
...
+// All additions are optional; existing serialised data remains parseable.
+// See docs/plans/MULTI_LAYER_AUDIO_FORMAT.md for the design rationale.
```
ashesfall: Do not put references to temporary documents in files like this.

PR #277, opened 2026-05-31
Metal sustained dispatch

ashesfall:


base/hardware/src/main/java/org/almostrealism/hardware/cl/CLDataContext.java
```diff
@@ -389,7 +389,7 @@ protected void identifyDevices(boolean kernelQueue) {
 				CL.clGetDeviceIDs(platform, CL.CL_DEVICE_TYPE_GPU, numDevices, devices, null);
 				kernelDevice = devices[deviceIndex];
 
-				log("Hardware[" + name + "]: Using " + deviceName(CL.CL_DEVICE_TYPE_GPU) + " " + deviceIndex + " for kernels");
```
ashesfall, line 392: Do not change the language here.

base/code/src/main/java/io/almostrealism/code/ComputeContext.java
```diff
@@ -120,6 +120,26 @@ default boolean isProfiling() {
...
+	 * has already completed the work by the time its {@code accept} returns) must return
+	 * {@code false} here: chaining such a provider would only thread its host-readiness latch
+	 * into the next operation's wait, serializing across the async arg-loading executor with no
+	 * benefit. Composites (see {@code OperationList}) chain only the operations whose context
+	 * reports {@code true} and run the rest sequentially.</p>
+	 *
+	 * @return {@code true} if dispatches here yield a deferrable device completion worth
+	 *         chaining; {@code false} (the default) for synchronous providers
+	 */
+	default boolean isCompletionDeferred() {
```
ashesfall: I don't mind this method existing, but it is always risk to expose things like this because it allows for special casing that is often counter productive. The contract for computation requires that everyone comply with the wait-on-Semaphore pattern - we do not want to encourage disregard for this pattern. It is essential that we review uses of this method and ensure that it is not being used as a way to avoid full compliance with our Semaphore pattern.

base/hardware/src/main/java/org/almostrealism/hardware/jni/NativeDataContext.java
```diff
@@ -239,6 +239,7 @@ public void init() {
 		}
 
 		context = isExternal ? new ExternalComputeContext(this, compiler) : new NativeComputeContext(this, compiler);
+		log("Hardware[" + getName() + "]: kernels run on the CPU");
```
ashesfall: Please use language that matches the original language you removed from CL compute context ("Using ... for ..")

base/hardware/src/main/java/org/almostrealism/hardware/mem/AcceleratedProcessDetails.java
```diff
@@ -232,6 +293,17 @@ public AcceleratedProcessDetails(Object[] args, int kernelSize,
 	 */
 	public void setSemaphore(DefaultLatchSemaphore semaphore) { this.semaphore = semaphore; }
```
ashesfall: Why are there two semaphores? There is no possible justification for this, so I assume a deliberate attempt to conceal failure to properly to complete the work.

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalCommandRunner.java
```diff
@@ -16,93 +16,194 @@
...
-	private MTLBuffer offset;
-	/** Pre-allocated integer buffer holding the element count (size) of each kernel argument. */
-	private MTLBuffer size;
 	/** The command queue used to submit encoded Metal compute commands. */
 	private final MTLCommandQueue queue;
 
+	/** The open (encoded, not yet committed) command buffer, or {@code null}. Executor-thread only. */
+	private MTLCommandBuffer openBuffer;
+	/** Autorelease pool spanning the open buffer's lifetime. Executor-thread only. */
+	private long openPool;
```
ashesfall: If this is a memory address for a native object, you must create a Java object to represent it. We are NOT doing this crap where we just stop caring about types just because we are using JNI. Whatever made you think that was okay, go fix the instructions (probably base/CLAUDE.md ?) to make it clear that won't be tolerated.

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalCommandRunner.java
```diff
@@ -16,93 +16,194 @@
...
+	}
+
+	/**
+	 * Returns the shared completion semaphore whose {@link Semaphore#waitFor()} commits the
+	 * open batch. Used as an operation's completion handle so the trailing wait of a chained
+	 * group triggers the single commit.
+	 *
+	 * @return the completion semaphore
+	 */
+	public Semaphore completionSemaphore() {
```
ashesfall: Why are you qualifying this with "completion"? Is there more than one semaphore?

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalDataContext.java
```diff
@@ -124,6 +124,7 @@ protected void identifyDevices() {
 		if (mainDevice != null) return;
 
 		mainDevice = MTLDevice.createSystemDefaultDevice();
+		log("Hardware[" + getName() + "]: kernels run on the system default GPU");
```
ashesfall: Please use the language that matches our original CL logging.

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalOperator.java
```diff
@@ -214,7 +212,11 @@ public synchronized Semaphore accept(Object[] args, Semaphore dependsOn) {
 
 		long id = totalInvocations++;
 
-		if (dependsOn != null) dependsOn.waitFor();
+		// When batching is off, each command commits and waits immediately, so a prior
+		// dependency must be satisfied before this dispatch. When batching is on, ordering is
+		// provided by the shared command buffer (one encoder per command, Metal hazard
+		// tracking) and the serial command queue, so no host-side dependsOn wait is needed.
+		if (!MetalCommandRunner.enableBatching && dependsOn != null) dependsOn.waitFor();
```
ashesfall: Why check the enable batching flag? if someone TOLD you that you depend on this semaphore, and you are REFUSING to comply - what do we call? what is the name for a behavior like that?

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalOperator.java
```diff
@@ -223,79 +225,64 @@ public synchronized Semaphore accept(Object[] args, Semaphore dependsOn) {
...
-		}
-
-		Reference.reachabilityFence(data);
-		Reference.reachabilityFence(args);
+			Reference.reachabilityFence(data);
+			Reference.reachabilityFence(args);
+		});
 
-		return null;
+		return MetalCommandRunner.enableBatching ? context.getCommandRunner().completionSemaphore() : null;
```
ashesfall: This is not the way to do this! If the command runner gives you a semaphore, YOU USE IT. If you think there should be no Semaphore returned, you should have the command runner be smart enough to not create one. These special case flags are ALWAYS a hazard. They ALWAYS encourage failure to comply with the design and I'm horrified to find how much worse that lack of compliance is than I even expected when I saw the carve outs in the first place.

base/hardware/src/main/java/org/almostrealism/hardware/OperationList.java
```diff
@@ -1292,17 +1318,51 @@ public void run() {
...
-					for (int i = 0; i < run.size(); i++) {
-						if (enableRunLogging)
-							log("Running " + OperationInfo.display(run.get(i)));
-						timingListener.recordDuration(getMetadata(), run.get(i));
+				Semaphore pending = null;
+
+				for (int i = 0; i < run.size(); i++) {
+					Runnable r = run.get(i);
+
+					if (enableSemaphoreChaining && r instanceof Submittable
```
ashesfall: So I just don't get batching if I don't use OperationList? That's our position, is that this is the only Computation anywhere in the entire platform that gets to use this feature?

base/hardware/src/main/java/org/almostrealism/hardware/OperationList.java
```diff
@@ -617,6 +619,30 @@ public class OperationList extends ArrayList<Supplier<Runnable>>
...
+	 * never chained.</p>
+	 *
+	 * <p><strong>Defaults on.</strong> The gate keeps it inert on synchronous providers; the only
+	 * provider that defers is batched Metal
+	 * ({@link io.almostrealism.code.ComputeContext#isCompletionDeferred()}). Dependent chained
+	 * members are encoded in order (see {@code AcceleratedOperation.submit()} ->
+	 * {@code AcceleratedProcessDetails.awaitReady()}), so ordering is correct. Set
+	 * {@code AR_HARDWARE_SEMAPHORE_CHAINING=disabled} to force the legacy per-operation wait.</p>
+	 */
+	public static boolean enableSemaphoreChaining =
```
ashesfall: This is absurd. ANOTHER special flag? You have got. to. be. kidding.

PR #296, opened 2026-06-11
AudioScene rendering DSP migrated to PDSL

ashesfall:


studio/compose/src/test/java/org/almostrealism/studio/ml/test/MixdownManagerPdslVerificationTest.java
```diff
@@ -290,6 +343,349 @@ public void compareJavaAndPdslMixdownPaths() throws IOException {
...
+	 * the delayed signal once without recirculating), identity passthrough, and a zero reverb
+	 * send, plus validly-shaped state buffers for every ring.
+	 *
+	 * @return argument entries to merge over the adapter-built map
+	 */
+	private Map<String, Object> neutralEfxArgs() {
+		int firTaps = PDSL_FILTER_ORDER + 1;
+		PackedCollection efxCoeffs = new PackedCollection(new TraversalPolicy(CHANNELS, firTaps));
+		PackedCollection efxWet = new PackedCollection(new TraversalPolicy(CHANNELS));
+		PackedCollection efxAuto = new PackedCollection(new TraversalPolicy(CHANNELS));
```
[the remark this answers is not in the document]
ashesfall: We verified this against every memory provider rather than assuming either way, because the guarantee has evolved over time. A plain `new PackedCollection(shape)` allocates directly from the active `MemoryProvider` (the plain constructor never goes through `Heap`), and that allocation is zero-initialized on every standard backend: `JVMMemoryProvider` backs with `new double[]` (zeroed by the JLS), `NativeMemoryProvider` emits `calloc` (see `Malloc.getFunctionDefinition`), `MetalMemoryProvider`'s default mode calls `newBufferWithLength` which the Metal API documents as a zero-filled allocation (`MTL.cpp` `createBuffer32`), and `NativeBufferMemoryProvider` uses `ByteBuffer.allocateDirect` (zeroed by the JVM spec). Providers don't pool freed chunks, so stale contents can't resurface through the plain constructor. The bare allocations here are therefore intentionally zero. The exceptions — OpenCL DEVICE/HOST locations, `Heap`-carved `factory()` allocations after stage reuse, and explicit shared-file scopes — don't apply to these test collections; we've documented the full table in `docs/internals/packed-collection-examples.md` ("Memory initialization guarantees") so this question has a durable answer.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/MixdownManagerPdslVerificationTest.java
```diff
@@ -290,6 +343,349 @@ public void compareJavaAndPdslMixdownPaths() throws IOException {
...
+		double[] ones = new double[CHANNELS];
+		Arrays.fill(ones, 1.0);
+		efxAuto.setMem(ones);
+
+		PackedCollection fbDelay = new PackedCollection(new TraversalPolicy(CHANNELS));
+		double[] fbDelayData = new double[CHANNELS];
+		Arrays.fill(fbDelayData, PDSL_DELAY_SAMPLES);
+		fbDelay.setMem(fbDelayData);
+		PackedCollection fbTransmission = new PackedCollection(new TraversalPolicy(CHANNELS, CHANNELS));
```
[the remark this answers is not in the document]
ashesfall: `fbTransmission` is guaranteed all-zeros as allocated: plain `PackedCollection` constructors are zero-initialized on every standard backend (JVM `new double[]`, JNI `calloc`, Metal `newBufferWithLength`, NIO `allocateDirect`) — verified per provider, with the full table now documented in `docs/internals/packed-collection-examples.md` under "Memory initialization guarantees". A zero transmission matrix is exactly the neutral no-feedback routing this smoke intends, so no change is needed.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/MixdownManagerPdslVerificationTest.java
```diff
@@ -290,6 +343,349 @@ public void compareJavaAndPdslMixdownPaths() throws IOException {
...
+		PackedCollection fbPassthrough = new PackedCollection(new TraversalPolicy(CHANNELS, CHANNELS));
+		double[] passthroughData = new double[CHANNELS * CHANNELS];
+		for (int i = 0; i < CHANNELS; i++) passthroughData[i * CHANNELS + i] = 1.0;
+		fbPassthrough.setMem(passthroughData);
+		PackedCollection fbBuffers = new PackedCollection(CHANNELS * PDSL_SIGNAL_SIZE);
+		PackedCollection fbHeads = new PackedCollection(CHANNELS);
```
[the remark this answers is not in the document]
ashesfall: `fbBuffers`/`fbHeads` start at zero by the provider allocation contract: plain `PackedCollection` constructors are zero-initialized on all standard backends (JVM `new double[]`, JNI `calloc`, Metal `newBufferWithLength`, NIO `allocateDirect`), verified per provider and now documented in `docs/internals/packed-collection-examples.md` ("Memory initialization guarantees"). The feedback recurrence therefore starts from clean ring state, and the suite's determinism is borne out empirically — the PDSL renders reproduce their windowed RMS to four decimal places across regenerations. No change needed.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/MixdownManagerPdslVerificationTest.java
```diff
@@ -290,6 +343,349 @@ public void compareJavaAndPdslMixdownPaths() throws IOException {
...
+
+		int reverbFrames = 4;
+		PackedCollection reverbSend = new PackedCollection(new TraversalPolicy(CHANNELS));
```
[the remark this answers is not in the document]
ashesfall: `reverb_send` is zero as allocated — plain `PackedCollection` constructors are zero-initialized on every standard backend (verified per provider: JVM `new double[]`, JNI `calloc`, Metal `newBufferWithLength`, NIO `allocateDirect`; see the new "Memory initialization guarantees" section in `docs/internals/packed-collection-examples.md`). A zero send is precisely the neutral reverb this smoke wants, so the bare allocation is intentional.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/MixdownManagerPdslVerificationTest.java
```diff
@@ -290,6 +343,349 @@ public void compareJavaAndPdslMixdownPaths() throws IOException {
...
+		PackedCollection reverbDelays = new PackedCollection(new TraversalPolicy(CHANNELS));
+		double[] reverbDelayData = new double[CHANNELS];
+		Arrays.fill(reverbDelayData, PDSL_SIGNAL_SIZE);
+		reverbDelays.setMem(reverbDelayData);
+		PackedCollection reverbFeedback = new PackedCollection(new TraversalPolicy(CHANNELS, CHANNELS));
```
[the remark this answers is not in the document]
ashesfall: `reverbFeedback` is the all-zeros matrix by the provider allocation contract (plain `PackedCollection` constructors are zero-initialized on all standard backends — verified per provider and documented in `docs/internals/packed-collection-examples.md`, "Memory initialization guarantees"). Zero feedback is a known-stable matrix — the network reduces to its feedforward taps — and with the zero `reverb_send` the arm contributes exactly nothing, which is what the neutral smoke asserts against. No change needed.

studio/compose/src/test/java/org/almostrealism/studio/ml/test/MixdownManagerPdslVerificationTest.java
```diff
@@ -290,6 +343,349 @@ public void compareJavaAndPdslMixdownPaths() throws IOException {
...
+		PackedCollection reverbBuffers = new PackedCollection(CHANNELS * reverbFrames * PDSL_SIGNAL_SIZE);
+		PackedCollection reverbHeads = new PackedCollection(CHANNELS);
```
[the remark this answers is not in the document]
ashesfall: `reverb_buffers`/`reverb_heads` begin zeroed by the provider allocation contract — plain `PackedCollection` constructors are zero-initialized on every standard backend (JVM `new double[]`, JNI `calloc`, Metal `newBufferWithLength`, NIO `allocateDirect`); we verified each provider's implementation and recorded the table in `docs/internals/packed-collection-examples.md` ("Memory initialization guarantees"). The ring state is therefore deterministic from the first forward pass, which matches the suite's observed run-to-run reproducibility. No change needed.

studio/compose/src/main/java/org/almostrealism/studio/AudioSceneRealtimeRunner.java
```diff
@@ -0,0 +1,395 @@
...
+	public TemporalCellular create(MultiChannelAudioOutput output,
+								   List<Integer> channels, int bufferSize) {
+		List<Integer> resolved = channels != null ? channels :
+				IntStream.range(0, scene.getChannelCount()).boxed().collect(Collectors.toList());
+
+		if (MixdownManager.enablePdslMixdown) {
+			return createPdsl(output, resolved, bufferSize);
+		}
+
+		return createCellList(output, resolved, bufferSize);
```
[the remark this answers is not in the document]
ashesfall: Good catch — implemented as suggested. `AudioSceneRealtimeRunner.create` now routes through a `supportsPdsl(channels)` guard: the PDSL path is taken only for selections of at least two channels forming the zero-based contiguous prefix [0..n-1] (the adapter's `concat`-based per-channel argument construction needs ≥2 inputs, and its gene reads are positional, so any other selection would mis-map genome/automation). Everything else — including the single-channel selection from `AudioScene.renderChannel` — falls back to the CellList runner with a log line stating why. Verified with the plumbing smoke (both paths render finite, non-silent output).

base/hardware/docs/INSTRUCTION_CACHING.md
```diff
@@ -53,33 +53,26 @@ Two operations are "structurally identical" if they perform the same computation
...
 |                                                           |
 +-----------------------------------------------------------+
 ```
 
-The `instructionsCache` is the primary cache for compiled kernels. It maps signature strings to `ScopeInstructionsManager` instances, each of which lazily compiles and caches an `InstructionSet`.
+The `instructionsCache` maps cache keys to `ScopeInstructionsManager` instances, each of which lazily compiles and caches an `InstructionSet`.
+
+**The cache key combines the computation signature with the identity of the `ComputeContext`.** A compiled kernel is permanently bound to the context it was compiled under — its operator dispatches through that context's command runner. If a structurally-identical operation from a *different* context were to adopt the kernel via a bare signature match, its dispatches would be encoded into a command runner that nothing in its own pipeline ever commits: the kernel would never execute and the operation would silently produce nothing (observed as exactly-zero output). Scoping the key per context preserves reuse within a context — the common case — while making cross-context structural twins compile independently.
```
ashesfall: The instructions cache is intentionally global and one of two things is true: (a) this is possible to achieve (if I planned to use ComputeContext A, but I find ComputeContext B has a cached copy of the operation I want, I should be able to *just use ComputeContext B*) and the changes here are a big mistake OR (b) the cache should be a property of the ComputeContext not the Computer.

I am *basically certain* the answer is (a)

base/hardware/src/main/java/org/almostrealism/hardware/arguments/ProcessArgumentMap.java
```diff
@@ -311,6 +323,93 @@ public void put(ProcessTreePositionKey key, Producer producer) {
...
+	/**
+	 * Logs every kernel argument with its position coverage: whether a tree position was
+	 * recorded for it (making it substitutable on reuse) and what its original producer
+	 * is. Arguments without positions keep their original (first operation's) binding
+	 * when the kernel is reused, which is exactly the state this diagnostic exposes.
+	 */
+	public void logCoverage() {
+		for (ArrayVariable<?> argument : arguments) {
+			if (argument == null) continue;
+			ProcessTreePositionKey position = positionsForArguments.get(argument);
```
ashesfall: Is ProcessTreePositionKey used at all? We should check if the test suite contains any examples of populating positionsForArguments, or if there is any way for this to happen, because I believe this is total superseded by our new method of caching InstructionSets which does not require remapping the tree by position key. We should be removing positionsForArguments, or maybe this entire ProcessArgumentMap class, not adding logging to it.

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedComputationOperation.java
```diff
@@ -377,6 +378,12 @@ public <K extends ExecutionKey> ComputableInstructionSetManager<K> getInstructio
 			if (ScopeSettings.enableInstructionSetReuse && signature != null) {
 				DefaultComputer computer = Hardware.getLocalHardware().getComputer();
 
+				if (ProcessArgumentMap.enableSubstitutionLogging) {
```
ashesfall: Again, I believe that this is not used - we should not be adding logging, we should be removing the outdated code.

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedComputationOperation.java
```diff
@@ -524,6 +548,23 @@ public Execution load() {
 		return operator;
 	}
 
+	/**
+	 * Logs the bindings of this operation's own {@link Computation} inputs
+	 * (index 0 is the destination for assignment-style operations), allowing
+	 * substituted argument bindings to be compared against the operation's
+	 * true producers when diagnosing instruction reuse.
+	 */
+	private void logOwnInputs() {
```
ashesfall: This is redundant with respect to Computation::describe for most Computation implementations - if calling describe on the Computation doesn't do what's needed, it should be improved; there should not be new methods for "alternative descriptions"

base/hardware/src/main/java/org/almostrealism/hardware/DefaultComputer.java
```diff
@@ -584,16 +524,25 @@ public ScopeInstructionsManager<ScopeSignatureExecutionKey> getScopeInstructions
 																							Computation<?> computation,
 																							ComputeContext<?> context,
 																							Supplier<Scope<?>> scope) {
+		// The cache key must include the ComputeContext: a compiled kernel is permanently
+		// bound to the context it was compiled under (its command runner dispatches the
+		// kernel), so an operation from another context that adopts it via a bare signature
+		// match would encode commands into a runner that nothing in its own pipeline ever
+		// commits - the kernel never executes and the operation silently produces nothing.
+		// Reuse is therefore scoped to operations that share both structure and context.
+		String cacheKey = Objects.requireNonNull(signature)
```
ashesfall: This is a big mistake, as described in an earlier comment.

compute/algebra/src/main/java/org/almostrealism/collect/computations/CollectionZerosComputation.java
```diff
@@ -193,21 +193,26 @@ public CollectionProducerParallelProcess generate(List<Process<?, ?>> children)
...
-	 *                                       applicable to zero computations
+	 * @return this instance, as a constant requires no isolation
 	 *
 	 * @see Process#isolate()
+	 * @see SingleConstantComputation#isolate()
 	 */
 	@Override
 	public Process<Process<?, ?>, Evaluable<? extends PackedCollection>> isolate() {
-		throw new UnsupportedOperationException();
+		return this;
```
ashesfall: This should never be isolated. In every case where you consider isolating this Computation, that means that it has at least one parent. And the set of parent Computations which can benefit from the overhead of separately computing the value 0.0 as a PackedCollection filled with zeros rather than an inline constant is the empty set. Find me *one* Computation for which there is a benefit here, and this could be made permissible, but there is no such Computation. Even trying to invent a useless one, simply to prove that one _could_ exist is a difficult conceptual problem.

compute/time/src/main/java/org/almostrealism/time/computations/MultiOrderFilter.java
```diff
@@ -327,4 +403,26 @@ public static MultiOrderFilter create(Producer<PackedCollection> series, Produce
...
+	 * {@code channels} contiguous rows, each convolved independently (no bleed across
+	 * row boundaries) with its own row of the {@code [channels, taps]} coefficient
+	 * bank, in a single kernel.
+	 *
+	 * @param series Producer providing {@code channels} rows of signal
+	 * @param coefficients Producer providing a {@code [channels, taps]} coefficient bank
+	 * @param channels Number of independent rows in the series
+	 * @return A new MultiOrderFilter instance ready for evaluation
+	 */
+	public static MultiOrderFilter createMultiChannel(Producer<PackedCollection> series,
```
ashesfall: There does not need to be two create methods, this is totally redundant and channels can be easily inferred from the shape. Remove this method.

engine/ml/src/main/java/org/almostrealism/ml/dsl/PdslBuiltins.java
```diff
@@ -0,0 +1,569 @@
...
+				"transformer() expects 19 arguments, got " + args.size());
+	}
+
+	/**
+	 * Builds a feed-forward (MLP) block from 4 or 5 evaluated weight arguments.
+	 *
+	 * @param args Evaluated arguments: RMSNorm weight, w1, w2, w3, and optionally epsilon
+	 * @return A feed-forward {@link Block}
+	 */
+	private static Object callFeedForward(List<Object> args) {
```
ashesfall: Never use Object, use a generic argument <T>. This applies all over the place in these PDSL classes; type handling is absolutely garbage in these classes and needs serious attention.

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManagerPdslAdapter.java
```diff
@@ -218,9 +221,462 @@ public static Map<String, Object> buildArgsMap(MixdownManager manager, Config co
...
+	 * copies of the two previous subtrees).
+	 *
+	 * @param high       true for the high-pass design; false for low-pass
+	 * @param cutoff     cutoff frequency in Hz (already bounded to [10, 20000])
+	 * @param sampleRate audio sample rate in Hz
+	 * @param out        destination array
+	 * @param offset     index of the first coefficient within {@code out}
+	 * @param count      number of taps to write
+	 */
+	private static void biquadImpulseResponse(boolean high, double cutoff,
```
ashesfall: This needs to be done with ComputationProducers, not java math.

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManagerPdslAdapter.java
ashesfall, line 71: Mark this class as deprecated with a note explaining that once PDSL is the standard form, this adapter could should be cleared away and the MixdownManager itself should be the "adapter".

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManagerPdslAdapter.java
```diff
@@ -218,9 +229,486 @@ public static Map<String, Object> buildArgsMap(MixdownManager manager, Config co
...
+	}
+
+	/**
+	 * Builds a row-major {@code [n, n]} diagonal matrix slot with {@code value} on the diagonal.
+	 *
+	 * @param n     matrix dimension
+	 * @param value the diagonal value
+	 * @return a {@code [n, n]} {@link PackedCollection} diagonal matrix
+	 */
+	private static PackedCollection diagonalMatrix(int n, double value) {
```
ashesfall, line 627: There is a specific CollectionProducer for this already, and it should be used.

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManagerPdslAdapter.java
ashesfall: Done in 46e266ea4. The class is marked `@Deprecated` with a `@deprecated` note: once PDSL is the standard mixdown path the adapter is cleared away and `MixdownManager` itself becomes the adapter (building its own argument map directly rather than routing through this glue).

studio/compose/src/main/java/org/almostrealism/studio/arrange/MixdownManagerPdslAdapter.java
ashesfall: Done in 46e266ea4 — full producer rewrite. `biquadResponseTable` is now a pure `CollectionProducer` closed-form and the Java IIR recurrence (`biquadImpulseResponse`) is deleted.

Because the biquad poles are always a complex-conjugate pair for the project resonance (`r=0.2`, the discriminant reduces to `-c²(4-r²) < 0`), the impulse response has the closed form `ρⁿ·(A·cos nθ + B·sin nθ)` with `ρ=√b2`, `θ=acos(-b1/2ρ)`, `y[0]=a1` — so the whole `[1024, taps]` table is built from per-bin scalar producers with no recurrence.

This needed a first-class `acos`, which didn't exist, so I added it end-to-end: `ArcCosine` expression node (with `delta`), `ExpressionArithmetic.acos()`, `ExpressionFeatures.acos(...)`, `GeometryFeatures.acos(...)`, and `using metal::acos;` in the Metal prelude. The table is materialized at the build boundary via an assignment op (`.get().run()`), not `.evaluate()`.

Validated element-wise against the old recurrence: max abs diff 9.0e-6 (HP) / 5.5e-6 (LP) across all 1024×41 (`BiquadResponseTableParityTest`), plus new `acos` forward + gradient unit tests.

engine/ml/src/main/java/org/almostrealism/ml/dsl/PdslBuiltins.java
ashesfall: Partially addressed. The `call*` dispatch methods now declare their real return types (`Function<TraversalPolicy, Block>`, `Function<TraversalPolicy, CellularLayer>`, `Block`, `TraversalPolicy`, `PackedCollection`) instead of `Object`; only the top-level `call()` dispatch entry and the genuinely-dual `callReshape` remain `Object`.

The broader generic-`<T>` redesign of dispatch and argument typing across these PDSL classes is a larger refactor I've kept as a dedicated follow-up rather than expanding this PR — agreed the type handling here needs serious attention.

compute/time/src/main/java/org/almostrealism/time/computations/MultiOrderFilter.java
ashesfall: Done — `createMultiChannel` is removed. The single `create(series, coefficients)` now infers the channel count from the coefficients' shape (`dimensions <= 1 ? 1 : length(0)`).

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedComputationOperation.java
ashesfall: Removed — the alternative-description method this branch added here is gone; descriptions go through `Computation::describe` as you suggested (reverted to master's version of this file).

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedComputationOperation.java
ashesfall: Logging removed (reverted to master). On "not used": the reuse path here is still live — `load()` calls `ProcessArgumentMap.putSubstitutions(...)` (~line 515) and only emits the `warn` at 518 when reuse isn't possible. This is the path that prevents the silent-zero reuse bug, so I removed the diagnostics but kept the functional code. More detail on `positionsForArguments` in the `ProcessArgumentMap` thread.

base/hardware/src/main/java/org/almostrealism/hardware/arguments/ProcessArgumentMap.java
ashesfall: Diagnostic logging removed. On whether it's used: `ProcessArgumentMap.putSubstitutions(...)` is still invoked on the instruction-reuse path at `AcceleratedComputationOperation#load` (~line 515), which populates `positionsForArguments`; those substitutions are what let a reusing op adopt a cached scope's arguments by tree position. When two structurally-identical models hash to the same signature, this is what rebinds the second op's producers to the right buffers — without it the earlier silent-zero reuse bug returns (second model produced exactly-zero output). `ProcessTreePositionKey` is also still an `ExecutionKey` used by `InstructionSetManager`.

So it isn't dead today. Fully removing `positionsForArguments` / `ProcessArgumentMap` would mean reworking that reuse path, which I'd suggest doing as its own change rather than folding into this PR — happy to scope it if you'd like.

PR #322, opened 2026-06-28
Streaming pattern element rendering

ashesfall:


studio/compose/src/main/java/org/almostrealism/studio/AudioSceneRealtimeRunner.java
```diff
@@ -438,9 +535,17 @@ private Map<String, Object> buildMixdownArgs(boolean wetVoicing, MixdownManager
 	private CompiledModel compileMixdownModel(PdslLoader loader, PdslNode.Program program,
 											  String layerName, TraversalPolicy inputShape,
 											  Map<String, Object> args) {
+		// The mixdown model is inference-only: every tick calls forward() over the whole buffer and
```
ashesfall: This class is heavily overusing inline comments. Key information and design detail should appear in class or method javadoc documentation, which remains searchable; inline comments should only be used for brief notes about internal implementation details.

studio/compose/src/main/java/org/almostrealism/studio/PatternRenderStream.java
```diff
@@ -0,0 +1,252 @@
...
+ * it: the producer waits on {@code empty} before rendering into the next slot and signals
+ * {@code filled} when done; the consumer waits on {@code filled}, mixes the slot, and signals
+ * {@code empty} on release. The producer never runs more than {@code slots} buffers ahead
+ * (back-pressure); the consumer blocks only if the producer falls behind (under-run).</p>
+ *
+ * <p>The producer advances its own {@code renderFrame} (the render-ahead position in frames),
+ * which is distinct from the consumer's playback clock. The render depends only on the frame
+ * parameter, so the two positions are free to diverge by up to {@code slots} buffers.</p>
+ */
+class PatternRenderStream implements Destroyable {
```
ashesfall: Classes should be public by default, unless you actively need to hide them.

studio/compose/src/main/java/org/almostrealism/studio/PatternRenderStream.java
```diff
@@ -0,0 +1,252 @@
...
+			while (running) {
+				empty.acquire();
+				if (!running) {
+					break;
+				}
+
+				long w = writeIndex.get();
+				renderFrame[0] = w * (long) bufferSize;
+				renderOp.run();
+				ring.setMem((int) (w % slots) * sliceSize, workingInput, 0, sliceSize);
```
ashesfall: This should use Assignment.

studio/compose/src/main/java/org/almostrealism/studio/PatternRenderStream.java
```diff
@@ -0,0 +1,252 @@
...
+	}
+
+	/**
+	 * Returns a view of the oldest filled ring slot, blocking until the producer has rendered
+	 * it. The caller must mix the returned slot and then call {@link #release()}. The view stays
+	 * valid until {@code release()} frees the slot for reuse.
+	 *
+	 * @return a {@code [inputChannels, bufferSize]} view of the next buffer's rendered input
+	 */
+	PackedCollection awaitSlot() {
```
ashesfall: This is surely part of the public interface of this type. Do not conceal methods unless they are intended to be hidden.

studio/music/src/main/java/org/almostrealism/music/pattern/PatternAudioBuffer.java
```diff
@@ -181,7 +196,7 @@ public Supplier<Runnable> prepareBatch() {
 		return () -> {
```
ashesfall: This should return an OperationList, which includes the sumOp and uses Assignment to zero out the outputBuffer.

studio/music/src/main/java/org/almostrealism/music/pattern/BatchedNoteInputs.java
```diff
@@ -69,28 +69,38 @@ public class BatchedNoteInputs {
...
-	/** Filter-envelope scalars {@code [5]}. */
+	/** Filter-envelope scalars {@code [5]}; {@code null} for percussion. */
 	private final double[] filterAdsr;
 
-	/** Volume-envelope scalars {@code [5]}. */
+	/** Volume-envelope scalars {@code [5]}; {@code null} for a dry percussion note. */
 	private final double[] volumeAdsr;
 
+	/** Whether this record is a percussion note (no per-layer envelope, no filter envelope, unity pitch). */
+	private final boolean percussion;
```
ashesfall: We use melodic as the flag everywhere, it does not make sense to invert that here unless this is intended to communicate something beyond just "not melodic"; if it is intended to communicate more than that, then a better term should be chosen that actually reflects that (eg, envelopes=false, etc)

studio/music/src/main/java/org/almostrealism/music/pattern/BatchedPatternLayerRenderer.java
```diff
@@ -433,6 +488,97 @@ private void dispatchWindow(PatternFeatures features, List<RenderedNoteAudio> no
...
+	 * placed, summed output into {@code destination}.
+	 *
+	 * @param features       the features instance providing the evaluate boundary
+	 * @param notes          the notes overlapping this sub-window (size {@code >= 1})
+	 * @param windowStart    the sub-window's absolute start frame
+	 * @param windowWidth    the sub-window's frame count (per-note row length)
+	 * @param destination    the per-tick destination buffer to accumulate into
+	 * @param destBaseOffset the frame offset into {@code destination} for this sub-window
+	 */
+	private void dispatchWindowPercussion(PatternFeatures features, List<RenderedNoteAudio> notes, int windowStart,
```
ashesfall: Why are you accepting PatternFeatures as an argument here? If you want to use the methods of PatternFeatures, the correct approach is to implement that interface. Please update the documentation of all XXXFeatures classes to be more clear about this.

ashesfall: Resolved. `BatchedPatternLayerRenderer` now implements `PatternFeatures` and the parameter is gone from `render` and the private dispatch methods; the per-note fallback (`renderNotes`) and the accumulate boundary come from the inherited defaults. Dropping the parameter made `render`'s signature identical to the `PatternFeatures.render` default, so it is now an explicit `@Override` — which is the right semantics here: for this class, rendering *is* the batched dispatch, with no flag check or delegation (`PatternFeatures.render` remains the flag-checking entry point for everything else).

For the documentation request: all 67 `*Features` interfaces now carry a standard paragraph in their class javadoc stating the mixin convention — implement the interface (the methods are stateless defaults) rather than accept or hold a `Features` instance. The two `*Features` types that are classes rather than interfaces (`ComposableAudioFeatures`, `PdslFeatures`) were left as-is since the statement doesn't apply to them.

Verified: full build green, sentinels/RMS tests pass, and the exact-peak parity gate is bit-identical (0.49411046504974365).

studio/music/src/main/java/org/almostrealism/music/pattern/BatchedNoteInputs.java
ashesfall: Resolved — the flag communicated nothing beyond "not melodic", so it now follows the convention: the field and accessor are `melodic`/`isMelodic()` (true for the full SSS shape with per-layer and filter envelopes; false for the percussion subset), with the two factory sites and the one dispatch call site flipped accordingly. The static shape classifiers (`isMelodicSssShape`/`isPercussionSssShape`) kept their names since they describe the specific note structures being detected rather than this record's flag. Sentinel and percussion-vs-per-note RMS tests pass.

studio/music/src/main/java/org/almostrealism/music/pattern/PatternAudioBuffer.java
ashesfall: Resolved. `prepareBatch(boolean)` now returns an `OperationList` directly: an `Assignment` zero of the output buffer (when `clearOutput`) followed by the `patterns.sum(...)` op — the lambda wrapper and the host-side `outputBuffer.clear()` are gone. The assignment is constructed as `a(traverseEach(cp(outputBuffer)), c(0.0))` so it takes the compiled-kernel path rather than the provider short-circuit (same construction as the renderer's batched source clears). Build laziness is unchanged since the list defers each child's `get()` to its own `get()`. Parity verified: the setup-breakdown exact-peak gate is bit-identical and `GenerateAudioFileTest` still renders at peak 1.00.

studio/compose/src/main/java/org/almostrealism/studio/PatternRenderStream.java
ashesfall: Resolved — nothing here was meant to be hidden. `awaitSlot()` is public, along with the rest of the type's API: the constructor, `start`, `release`, `buffersRendered`, `buffersConsumed`, and `stop`. Only `produceLoop` and the new `slotCopy` helper remain private (genuinely internal).

studio/compose/src/main/java/org/almostrealism/studio/PatternRenderStream.java
ashesfall: Resolved. The ring copy is now a compiled `Assignment` per slot — the ring depth is fixed, so each slot's copy (`a(traverseEach(cp(slotRange)), traverseEach(cp(workingInput)))`) compiles exactly once, lazily on the slot's first use, and is destroyed with the stream. The `traverseEach` wrapping keeps it on the kernel path (dodging the provider short-circuit, matching the `copy()` helpers' construction), so the copy is a device dispatch ordered against the render's own kernels instead of the host-staged `setMem` round trip. Consumer ordering is unchanged: the mixdown forward reads the slot as kernel input on the same compute context. Parity verified bit-identical on the setup-breakdown exact-peak gate.

studio/compose/src/main/java/org/almostrealism/studio/PatternRenderStream.java
ashesfall: Resolved — the class is public (it also now implements `CollectionFeatures` for the assignment-based slot copies). The package-private default was concealment without intent; there is nothing about this type that needs hiding.

studio/compose/src/main/java/org/almostrealism/studio/AudioSceneRealtimeRunner.java
ashesfall: Resolved. The design detail moved into searchable javadoc: `createPdsl`'s method javadoc gained Wiring, Render-ahead, Per-buffer automation, and Output-streaming sections (covering the consolidated-buffer layout and layer choice, the dual-mono rationale, the channel-index genome mapping, the once-per-round consolidated zero, the automation-slot refresh, and the stable-output-handle capture); the anonymous `TemporalCellular` overrides in both runners now carry their own javadoc (setup's pre-warm/prefill semantics, tick's hot-path contract and clock-advance rationale, reset's clock-rewind rationale); and `compileMixdownModel`'s inference-only `compile(false)` rationale is in its javadoc. Remaining inline comments are one-line implementation notes.

PR #329, opened 2026-06-30
Introduce ComputeContext copy operation

ashesfall:


base/hardware/src/main/java/org/almostrealism/hardware/computations/Assignment.java
```diff
@@ -396,6 +445,41 @@ public Runnable get() {
...
+	 * kernel decision — to {@code context} rather than a compiler-selected one.
+	 *
+	 * <p>This is how {@link org.almostrealism.hardware.mem.MemoryDataArgumentMap} makes an aggregate
+	 * copy run under the {@link ComputeContext} of the {@link AcceleratedOperation} whose arguments it
+	 * prepares, so the copy and the kernel program it feeds share a single context.</p>
+	 *
+	 * @param context the ComputeContext to compile against, or null to select one as {@link #get()} does
+	 * @return the compiled runnable
+	 */
+	public Runnable get(ComputeContext<MemoryData> context) {
```
ashesfall: The compute context field should be final, you cannot change it like this that is incredibly dangerous. You have to set this stuff up at construction time.

base/hardware/src/main/java/org/almostrealism/hardware/computations/Assignment.java
```diff
@@ -190,6 +199,15 @@ public class Assignment<T extends MemoryData> extends OperationComputationAdapte
...
 	private final int memLength;
 
+	/**
+	 * The {@link ComputeContext} to compile this assignment's kernel against, or {@code null} to let
+	 * the compiler select one. Set via {@link #get(ComputeContext)} by the {@link AcceleratedOperation}
+	 * whose argument preparation this assignment performs, so the copy — and therefore {@link Runner}'s
+	 * direct-copy-vs-kernel decision — runs under the SAME context as the kernel program it feeds,
+	 * rather than an independently selected one.
+	 */
+	private ComputeContext<MemoryData> computeContext;
```
ashesfall: The parent class contains this already, do not duplicate it.

base/code/src/main/java/io/almostrealism/code/ComputeContext.java
```diff
@@ -110,6 +111,34 @@ public interface ComputeContext<MEM> {
...
+	 * never confused with a copy to another.</p>
+	 *
+	 * @param source            the memory to copy from
+	 * @param sourceOffset      the element offset within {@code source} to copy from
+	 * @param destination       the memory to copy into
+	 * @param destinationOffset the element offset within {@code destination} to copy into
+	 * @param length            the number of elements to copy
+	 * @return the copy's completion semaphore, or {@code null} when the copy completed synchronously
+	 */
+	default Semaphore copy(Memory source, int sourceOffset, Memory destination, int destinationOffset, int length) {
```
ashesfall: This should operate on MemoryData (which includes the offsets and lengths) rather than directly on Memory.

base/code/src/main/java/io/almostrealism/concurrent/Semaphore.java
```diff
@@ -48,6 +48,25 @@ public interface Semaphore {
...
+	 * <p>Two semaphores whose batch tokens are both non-null and not {@link Object#equals equal}
+	 * were encoded into different batches. This lets a caller that submitted a group of dispatches
+	 * which must share a batch — so the provider's in-batch hazard tracking can order their
+	 * read-after-write dependencies — detect after the fact that the group was split across batches
+	 * (e.g. because it exceeded the provider's per-batch limit), in which case that ordering is no
+	 * longer guaranteed.</p>
+	 *
+	 * @return an opaque batch token, or {@code null} if the provider exposes no batch grouping
+	 */
+	default Object getBatch() {
```
ashesfall: If we need this, there should be a dedicated class for representing it in a general way. Methods that return Object are not an option.

base/hardware/src/main/java/org/almostrealism/hardware/computations/Assignment.java
```diff
@@ -479,4 +528,92 @@ public String signature() {
...
+			Semaphore complete = submit(null);
+			if (complete != null) complete.waitFor();
+		}
+
+		@Override
+		public Semaphore submit(Semaphore dependsOn) {
+			MemoryData dst = resolve(destination);
+			MemoryData src = resolve(source);
+
+			if (dependsOn != null) dependsOn.waitFor();
```
ashesfall: dependsOn should be passed to ComputeContext.copy, it should be up to the ComputeContext to decide how to handle sequencing when there is an existing Semaphore active.

base/hardware/src/main/java/org/almostrealism/hardware/computations/Assignment.java
```diff
@@ -479,4 +528,92 @@ public String signature() {
...
+	 * <p>A pure memory-to-memory copy — both the destination and the source are {@link Provider}s,
+	 * which is exactly the case {@link #get()} runs via {@link Runner} — has only two arguments and
+	 * gains nothing from aggregation; aggregating it would instead fold its own operands into a
+	 * nested aggregate and emit pointless copy kernels (and, for small operands, recurse). Such a
+	 * copy therefore opts out. Every other assignment (whose source is a real computation) aggregates
+	 * normally.</p>
+	 *
+	 * @return {@code false} for a pure Provider-to-Provider copy, {@code true} otherwise
+	 */
+	public boolean isArgumentAggregationSupported() {
```
ashesfall, line 546: Does this override something? It is missing an Overrides annotation.

base/hardware/src/main/java/org/almostrealism/hardware/mem/MemoryDataArgumentMap.java
```diff
@@ -125,9 +148,23 @@ public MemoryDataArgumentMap(ArgumentProvider delegateProvider) {
...
+
+	/**
+	 * Creates an argument map.
+	 *
+	 * @param delegateProvider the argument provider used to create new argument variables
+	 * @param context the {@link ComputeContext} of the operation this map prepares arguments for, or
+	 *                null to let the aggregate copies select a context when compiled
+	 * @param aggregateGenerator factory for the aggregate buffer, or null to disable aggregation
+	 */
+	public MemoryDataArgumentMap(ArgumentProvider delegateProvider, ComputeContext<MemoryData> context,
```
ashesfall: ComputeContext should be the first argument here. And why are we preserving methods which allow it to be null? It can never be null.

base/hardware/src/main/java/org/almostrealism/hardware/mem/MemoryDataArgumentMap.java
```diff
@@ -288,64 +325,117 @@ public String describeAggregate() {
...
+	 * @param source the memory to copy from
+	 * @param target the memory to copy into
+	 * @return a submittable copy operation
+	 */
+	private Submittable copyOperation(MemoryData source, MemoryData target) {
+		int length = source.getMemLength();
+
+		return dependsOn -> {
+			if (context != null) {
+				return context.copy(source.getMem(), source.getOffset(),
```
ashesfall: dependsOn needs to be passed into the copy method, so that the ComputeContext can decide how to handle sequencing when a Semaphore is active.

base/hardware/src/main/java/org/almostrealism/hardware/mem/MemoryDataArgumentMap.java
```diff
@@ -395,7 +503,7 @@ public static boolean isAggregateArgument(ArrayVariable<?> arg) {
 	 * @return Fully configured {@link MemoryDataArgumentMap}
 	 */
 	public static MemoryDataArgumentMap create() {
```
ashesfall: Who uses this method?

base/hardware/src/main/java/org/almostrealism/hardware/mem/MemoryDataArgumentMap.java
```diff
@@ -405,10 +513,24 @@ public static MemoryDataArgumentMap create() {
 	 * @return Fully configured {@link MemoryDataArgumentMap}
 	 */
 	public static MemoryDataArgumentMap create(IntFunction<MemoryData> aggregateGenerator) {
```
ashesfall: Who uses this method?

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalComputeContext.java
```diff
@@ -199,6 +203,38 @@ public synchronized InstructionSet deliver(Scope scope) {
...
+	 * @return the blit's completion semaphore, or {@code null} for the direct-copy fallback
+	 */
+	@Override
+	public Semaphore copy(Memory source, int sourceOffset, Memory destination, int destinationOffset, int length) {
+		if (source instanceof MetalMemory && destination instanceof MetalMemory) {
+			MetalMemory src = (MetalMemory) source;
+			MetalMemory dst = (MetalMemory) destination;
+			long elementSize = src.getProvider().getNumberSize();
+
+			return runner.submit(
```
ashesfall, line 236: It is not valid for this use of submit, which obviously has to respect the Semaphore that is currently in effect, to pass null for dependsOn. Once copy properly accepts dependsOn, it needs to be threaded through here.

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalComputeContext.java
```diff
@@ -199,6 +203,38 @@ public synchronized InstructionSet deliver(Scope scope) {
...
+					cmdBuf -> cmdBuf.blitCopy(src.getMem(), sourceOffset * elementSize,
+							dst.getMem(), destinationOffset * elementSize, length * elementSize),
+					null, null);
+		}
+
+		((MemoryProvider) destination.getProvider())
```
ashesfall: We are going to have to wait for the active Semaphore before doing this, clearly.

base/hardware/src/main/java/org/almostrealism/hardware/metal/MetalSemaphore.java
```diff
@@ -75,6 +75,16 @@ public MetalSemaphore(OperationMetadata requester, MetalCommandRunner runner,
...
 
+	/**
+	 * Returns the command buffer this dispatch was encoded into as the batch token. Dispatches that
+	 * accumulated into the same open command buffer share the same instance; a forced commit (e.g.
+	 * the per-buffer limit) starts a new buffer, so a later dispatch returns a different instance.
+	 *
+	 * @return the {@link MTLCommandBuffer} this dispatch was encoded into
+	 */
+	@Override
+	public Object getBatch() { return commandBuffer; }
```
ashesfall: This leaks something that should be private. The identifier of a batch should be a special object designed for identifying the batch, not just hack like this.

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedComputationOperation.java
```diff
@@ -308,6 +309,18 @@ public boolean isFixedCount() {
...
+	/**
+	 * {@inheritDoc}
+	 *
+	 * <p>Defers to the wrapped {@link Assignment} when present, so a pure memory-to-memory copy can
+	 * opt out of folding its two operands into a nested aggregate.</p>
+	 */
+	@Override
+	protected boolean isArgumentAggregationSupported() {
+		return !(getComputation() instanceof Assignment)
+				|| ((Assignment<?>) getComputation()).isArgumentAggregationSupported();
```
ashesfall: This is completely insane. If you need a general mechanism for marking things as supporting argument aggregation or not, it should be through an interface not a special case for the Assignment class.

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedOperation.java
```diff
@@ -563,66 +583,163 @@ protected synchronized AcceleratedProcessDetails apply(MemoryBank output, Object
...
+	 * context (e.g. Metal) queues each copy onto the same open command buffer as the kernel, where
+	 * in-buffer hazard tracking orders the copy after the kernel; the returned semaphore only signals
+	 * once that buffer has run, so it is a sound completion handle. A direct-copy context runs each copy
+	 * synchronously (the kernel is already complete by then). Either way, no explicit dependency is
+	 * needed — and passing one would force the batching context to commit (split) the buffer.</p>
+	 *
+	 * @param operations  the operations to submit, in order
+	 * @return the completion of the last submitted operation, or null when {@code operations} is empty
+	 */
+	protected static Semaphore submitBatched(List<Submittable> operations) {
```
ashesfall: This should be a static method of Submittable, not some special thing you only use once buried in this class.

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedOperation.java
```diff
@@ -563,66 +583,163 @@ protected synchronized AcceleratedProcessDetails apply(MemoryBank output, Object
 			// Prepare the operator
 			Execution operator = setupOperator(process);
 
-			// Two host-mediated memory mechanisms may wrap the kernel (see method javadoc):
-			// compile-time argument aggregation (aggregating) and cross-provider replacement
-			// (processing). They are set up here and unwound in reverse order after the kernel runs.
+			// Two memory mechanisms may wrap the kernel (see method javadoc): compile-time argument
```
ashesfall, line 568: A bunch of the documentation in this class is out of date, and you are overusing inline comments when you should describe design patterns in javadoc and reserve inline comments for brief explanations of the details.

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedOperation.java
```diff
@@ -302,12 +308,26 @@ protected void prepareScope() {
...
 
+	/**
+	 * Returns whether this operation's small input arguments may be folded into a single aggregate
+	 * kernel argument. Defaults to {@code true}; an operation whose arguments must not be aggregated
+	 * (for example a pure memory-to-memory copy, which has only two arguments and would otherwise
+	 * fold its own operands into a nested aggregate) overrides this to return {@code false}.
+	 *
+	 * @return {@code true} if argument aggregation is permitted for this operation
+	 */
+	protected boolean isArgumentAggregationSupported() {
```
ashesfall: This should be an interface method.

base/hardware/src/main/java/org/almostrealism/hardware/AcceleratedOperation.java
```diff
@@ -195,6 +195,12 @@ public abstract class AcceleratedOperation<T extends MemoryData> extends Operati
 	 */
 	protected MemoryDataArgumentMap argumentMap;
 
+	/**
+	 * Guards the aggregation batch-split warning so it is emitted at most once per operation rather
+	 * than on every dispatch that straddles a provider command-buffer boundary.
+	 */
+	private boolean aggregateBatchSplitWarned;
```
ashesfall: Don't do this.

base/hardware/src/main/java/org/almostrealism/hardware/computations/Assignment.java
```diff
@@ -379,10 +409,19 @@ public Runnable get() {
...
-			// DestinationEvaluable can be used for AccelerationOperations and Providers
-			boolean shortCircuit = ev instanceof AcceleratedOperation<?> || ev instanceof Provider<?>;
+			// When both the destination and the source are Providers, the assignment is a plain
+			// memory-to-memory copy: short-circuit to the ComputeContext's copy() (see Runner), which
+			// moves the memory however the context prefers (a direct setMem, through the destination's
+			// MemoryProvider, by default). No kernel is compiled, so there is nothing to reuse — the
+			// offsets are read from the resolved memories on every run, so a copy into one region is
+			// never confused with a copy into another.
+			if (ev instanceof Provider<?>) {
+				return new Runner(Hardware.getLocalHardware().getComputeContext(),
```
ashesfall, line 419: You can't just grab the default ComputeContext, there is a ComputeContext for this Assignment, and thats the one to use.

PR #330, opened 2026-07-02
Consolidate kernel program flow control on Semaphore model

ashesfall:


base/hardware/src/main/java/org/almostrealism/hardware/computations/Assignment.java
```diff
@@ -379,10 +409,19 @@ public Runnable get() {
...
-			// DestinationEvaluable can be used for AccelerationOperations and Providers
-			boolean shortCircuit = ev instanceof AcceleratedOperation<?> || ev instanceof Provider<?>;
+			// When both the destination and the source are Providers, the assignment is a plain
+			// memory-to-memory copy: short-circuit to the ComputeContext's copy() (see Runner), which
+			// moves the memory however the context prefers (a direct setMem, through the destination's
+			// MemoryProvider, by default). No kernel is compiled, so there is nothing to reuse — the
+			// offsets are read from the resolved memories on every run, so a copy into one region is
+			// never confused with a copy into another.
+			if (ev instanceof Provider<?>) {
+				return new Runner(Hardware.getLocalHardware().getComputeContext(),
```
ashesfall: You can't just use the default ComputeContext here - the target needs to be specifically the ComputeContext that this Assignment is going to be compiled with. There are a number of important heuristics to make this decision, but critically the logic should match DefaultComputer::compileRunnable when no other information is available. There are cases where other choices make sense, however: and the Runner that operates directly on MemoryData is one of them - we would clearly want to favor a ComputeContext associated with the DataContext that supports the MemoryProvider for the Memory of the destination MemoryData.

We don't know that at the time the Runner is created, so what we should favor instead is a pattern where the Runner identifies the DataContext (and from it, the ComputeContext to target) when it runs. Critically, however, we will eventually care about this ComputeContext field when Runner is expanded to cover other cases; so we should also correct it, ensuring that *we select the ComputeContext that DefaultComputer::compileRunnable would select* for consistency with the compiled-kernel Assignment pattern.

We should also put a comment on Hardware.getLocalHardware() that warns about attempting to use it to make guesses about what hardware systems to target rather than using the Computer instance which is the one that actually knows how to make these choices intelligently.

base/hardware/src/main/java/org/almostrealism/hardware/ctx/AbstractComputeContext.java
```diff
@@ -155,6 +156,29 @@ public void runLater(Runnable runnable) {
...
+	 * buffer or queue overrides this to move the memory asynchronously and honor {@code dependsOn}
+	 * without a host stall.</p>
+	 *
+	 * @param source      the memory region to copy from
+	 * @param destination the memory region to copy into
+	 * @param dependsOn   the completion this copy must be ordered after, or {@code null}
+	 * @return {@code null}; the copy is complete on return
+	 */
+	@Override
+	public Semaphore copy(MemoryData source, MemoryData destination, Semaphore dependsOn) {
```
ashesfall, line 175: It is strongly preferable that this just add an operation that "happens after" dependsOn, returning a new Semaphore rather than calling waitFor here. If we are not going to do this now, we should at least put a TODO note indicating that this would be the ideal. It is wrong to assume that we ever have to perform blocking operations without the end user explicitly demanding it, because even when there is no native mechanism for queuing an operation (as in this case, where we must call setMem directly) we always have the fallback option to just use regular Java threading to chain operations together.

base/hardware/src/main/java/org/almostrealism/hardware/MemoryData.java
```diff
@@ -755,6 +756,22 @@ default void setMem(int offset, MemoryData src) {
...
+	 * {@link io.almostrealism.code.ComputeContext}'s
+	 * {@link io.almostrealism.code.ComputeContext#copy(Object, Object, io.almostrealism.concurrent.Semaphore) copy},
+	 * waiting for the copy to complete before returning. The context chooses the copy
+	 * mechanism (a direct transfer through the destination's provider, or a batched
+	 * device-side copy), so this is the preferred way to move data between two resolved
+	 * {@link MemoryData} regions from a synchronous caller.
+	 *
+	 * @param source the memory to copy from
+	 */
+	default void copyFrom(MemoryData source) {
```
ashesfall: This should not exist. I'm not entirely sure what it is used for, but you cannot just assume that the default ComputeContext is safe for this. Your abuse of Hardware.getLocalHardware().getComputeContext() is convincing me that we probably need to remove the option to obtain a ComputeContext this way, since it creates a risk that you do something insane like grab the default ComputeContext (probably JNI) and use it to copy Metal memory (very, very, very stupid).

If this actually made sense, it should be the mechanism that MemoryData::setMem uses (this method is a duplicate). But given it doesn't make sense, it should be removed. If anything relies on this, we need to investigate why the code that uses it doesn't know how to use the existing mechanisms we have built (Assignment as the primary option, and the use of ComputeContext::copy *in the case where we actually know what ComputeContext is correct*)

base/hardware/src/main/java/org/almostrealism/hardware/OperationListRunner.java
```diff
@@ -0,0 +1,178 @@
...
+				Hardware.getLocalHardware().getComputer().pushRequirements(requirements);
+			}
+
+			Semaphore pending = null;
+
+			for (int i = 0; i < run.size(); i++) {
+				Runnable r = run.get(i);
+
+				if (r instanceof Submittable) {
+					// Issue the dispatch without waiting, chaining it on the previous member's
```
ashesfall: Please do not use inline comments this way. Describe key functionality in the searchable javadoc comments for methods, classes, etc. Only use inline comments for brief explanations of subtle details that the outside world does not need to know about, not long discussions about the design and function of the system. A rule of thumb is if it requires more than 3 lines, you've made a mistake.

base/hardware/src/main/java/org/almostrealism/hardware/OperationListRunner.java
```diff
@@ -0,0 +1,178 @@
...
+	public List<Runnable> getOperations() { return run; }
+
+	/**
+	 * Executes all operations in sequence with compute requirements and timing.
+	 */
+	@Override
+	public void run() {
+		try {
+			if (requirements != null) {
+				Hardware.getLocalHardware().getComputer().pushRequirements(requirements);
```
ashesfall, line 108: Now that everything is async, will these requirements be respected? Aren't they managed by a ThreadLocal?

base/hardware/src/main/java/org/almostrealism/hardware/computations/Assignment.java
```diff
@@ -379,10 +409,22 @@ public Runnable get() {
 			}
 
 
-			// DestinationEvaluable can be used for AccelerationOperations and Providers
-			boolean shortCircuit = ev instanceof AcceleratedOperation<?> || ev instanceof Provider<?>;
+			// When both the destination and the source are Providers, the assignment is a plain
```
ashesfall: Do not use inline comments excessively like this. Put design-level detail in the searchable javadoc comments at method or class level and reserve inline comments for brief, private details about implementation nuance.

base/hardware/src/main/java/org/almostrealism/hardware/mem/MemoryDataCopy.java
```diff
@@ -302,8 +312,20 @@ public Runnable get() {
...
-			// TODO  This can be done faster if the source and target are on the same MemoryProvider
-			target.setMem(targetPosition, source.toArray(sourcePosition, length));
+			ComputeContext<MemoryData> context =
+					Hardware.getLocalHardware().getComputeContext(target.getMem());
+
+			if (context == null) {
+				// The target's memory is not managed by any configured DataContext
+				// (plain JVM heap, for example), so no backend can move it; fall back
+				// to a host-mediated transfer.
+				target.setMem(targetPosition, source.toArray(sourcePosition, length));
```
ashesfall: This can be accomplished without using source.toArray (letting source position and length be arguments directly) and the performance of that for some MemoryProviders will be better.

base/hardware/src/main/java/org/almostrealism/hardware/DestinationEvaluable.java
```diff
@@ -341,6 +341,15 @@ public void request(Object[] args) {
 		if (operation instanceof AcceleratedOperation) {
 			AcceleratedProcessDetails details = ((AcceleratedOperation) operation)
 					.apply(destination, Stream.of(args).map(arg -> (MemoryData) arg).toArray(MemoryData[]::new));
+			// Wait for the dispatch to be issued before reading the completion: until the whenReady
```
ashesfall: Do not use inline comments excessively like this.

base/hardware/src/main/java/org/almostrealism/hardware/Hardware.java
```diff
@@ -1109,7 +1120,7 @@ public <T> T computeContext(Callable<T> exec, ComputeRequirement... expectations
 	/**
 	 * Returns the off-heap buffer size for the specified backend.
 	 *
-	 * <p>Controlled by {@code AR_HARDWARE_OFF_HEAP_SIZE} environment variable (default: 1024 bytes).</p>
+	 * <p>Controlled by {@code AR_HARDWARE_OFF_HEAP_SIZE} environment variable (default: 0 bytes).</p>
```
ashesfall: Let's make this a constant, so we can reference it using `@value` and not be constantly worried about documentation falling out of date with the code.

compute/algebra/src/main/java/org/almostrealism/collect/computations/CollectionProvider.java
```diff
@@ -142,7 +141,7 @@ public Multiple<T> createDestination(int size) {
 	 * Creates an {@link Evaluable} that efficiently copies the provided collection into
 	 * the specified destination buffer.
 	 *
-	 * <p>This method uses {@link MemoryDataCopy} to perform hardware-accelerated memory
+	 * <p>This method uses the ComputeContext copy to perform hardware-accelerated memory
```
ashesfall: There should not be changes to the documentation if there aren't matching changes to the code, unless the documentation was wrong to begin with which I don't believe is true.

PR #334, opened 2026-07-04
Remove relative value support and correct CL issues

ashesfall:


.github/workflows/analysis.yaml
```diff
@@ -1345,13 +1339,19 @@ jobs:
            -DAR_TEST_PROFILE=pipeline
 
       - name: Run Music Tests
+        # Run even when an earlier module's tests failed (but not when the job was
+        # cancelled) so this informational lane reports every module's CL behavior
+        # rather than stopping at the first failure. The job still fails overall.
+        if: success() || failure()
```
ashesfall: This isn't necessary - we only want to avoid failing other jobs, we do not need to continue running later steps of a failed job.

.github/workflows/analysis.yaml
```diff
@@ -1162,7 +1153,10 @@ jobs:
       group: test-cl-${{ github.ref }}-${{ matrix.group }}
       cancel-in-progress: true
     strategy:
-      fail-fast: true
+      # This lane is informational (not in the all-checks gate) and is being used
```
ashesfall: This comment is excessive, we do not need to narrate every single change.

PR #336, opened 2026-07-04
Eliminate AcceleratedOperation output argument

ashesfall:


docs/plans/DROP_OPERATION_OUTPUT_ARG.md
```diff
@@ -40,6 +40,57 @@ implementation, and enumerates the risks. It is a study, not an approved change.
 
 ---
 
+## Status
```
ashesfall, line 43: So, after reviewing this work I have come to the opposite conclusion: having the output argument be of a special kind, distinct from the other arguments, does provide value that justifies keeping it.

Even if we drop the destination factory machinery: (1) the output argument is an argument that isn’t passed to children (they have their own) and (2) the presence of an output argument allows us to save time by skipping a lot of aggregation copy-out operations.

These justify keeping it. We should pivot the plan to focus ONLY on the destination factory simplification and cleanup, while leaving the dedicated output argument intact.

PR #342, opened 2026-07-08
Promote local variables in Repeated scopes and cache CLMemory on the JVM heap

ashesfall:


base/code/src/main/java/io/almostrealism/expression/Expression.java
```diff
@@ -585,6 +585,34 @@ public boolean containsStaticReferenceToAny(Set<String> names) {
...
+	 * whose referent {@link Variable} name matches one of the given names.
+	 *
+	 * <p>This is useful for detecting reads or writes of specific array variables,
+	 * which appear as {@link InstanceReference} nodes rather than
+	 * {@link StaticReference} nodes.</p>
+	 *
+	 * @param names the set of variable names to look for
+	 * @return true if a matching {@link InstanceReference} is found anywhere in this subtree
+	 */
+	public boolean containsInstanceReferenceToAny(Set<String> names) {
```
ashesfall: This method name is unnecessarily long, it should be containsReference.

We also should not have so many implementations of containsReference: the containsReference method which accepts a Variable should be preferred (we can accept a Set of Variables too), but if it is prohibitive for users to use Variable then the name option can be introduced but it should work *in the way* that containsReference does: a default that delegates to a children, and implementations directly in the subclasses that handle the various cases.

This is another serious violation of object oriented principles: you are packing all of the complexity into this parent class rather than embracing the flexibility that is provided by having a type hierarchy in the first place. With methods like this, we might as well just have a single Expression class and let it be in one of many modes. 

**Stop ignoring software design.**

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -88,6 +92,28 @@ public class Repeated<T> extends Scope<T> {
...
+	 * testing (comparing promotion-enabled vs promotion-disabled output).
+	 * Production code should not disable promotion.</p>
+	 *
+	 * <p>The transform assumes distinct kernel argument arrays do not overlap in
+	 * memory, which is the same assumption made throughout scope compilation. It
+	 * is conservative in every other respect: it bails out on nested loops,
+	 * method calls, metrics, non-assignment statements, and any reference to the
+	 * promoted array that is not structurally identical to the promoted element.</p>
+	 */
+	public static boolean enableAccumulatorPromotion = true;
```
ashesfall: This belongs in ScopeSettings or it should literally be removed (hopefully the latter).

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -97,6 +123,9 @@ public class Repeated<T> extends Scope<T> {
 	/** The boolean guard expression evaluated before each iteration; the loop continues while true. */
 	private Expression<Boolean> condition;
 
+	/** Statements rendered after the loop closes (e.g. final accumulator stores). */
+	private List<Statement<?>> epilogue = new ArrayList<>();
```
ashesfall: If you want to render something after the loop, you should do it in the structurally correct manner:

Repeated converts to a Scope that contains a new Repeated (the loop) followed by another Scope (containing the statements that come after).

You are breaking the pattern for your special case.

**Stop ignoring software design.**

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -192,6 +221,14 @@ public Repeated(Variable<Integer, ?> idx, Expression<Boolean> condition, Express
 	 */
 	public void setCondition(Expression<Boolean> condition) { this.condition = condition; }
 
+	/**
+	 * Returns the statements rendered after the loop closes, such as the final
+	 * accumulator stores produced by accumulator promotion.
+	 *
+	 * @return the epilogue statements
+	 */
+	public List<Statement<?>> getEpilogue() { return epilogue; }
```
ashesfall: No.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -204,6 +241,7 @@ public void write(CodePrintWriter w) {
 		w.println(getIndex().getName() + " = " + getIndex().getName() + " + " + interval.getExpression(w.getLanguage()) + ";");
 		w.println("}");
 
+		for (Statement s : getEpilogue()) { w.println(s); }
```
ashesfall: No

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -227,6 +265,10 @@ public Scope<T> simplify(KernelStructureContext context, int depth) {
 			hoistLoopInvariantStatements(scope);
 		}
 
+		if (enableAccumulatorPromotion && index != null && index.getName() != null) {
+			promoteAccumulators(scope);
```
ashesfall: This should not mutate the Scope, it should create a new Scope.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	 * <p>Promotion of an array is abandoned unless every reference to that array
+	 * in the loop body (read or write, in any statement) is structurally identical
+	 * to one of the promoted element references, and the loop condition and
+	 * interval do not reference the array at all. Pre-loop statements are left
+	 * untouched: the local is initialized from a load of the element, so any
+	 * initialization store that precedes the loop is still observed.</p>
+	 *
+	 * @param scope the simplified scope to optimize
+	 */
+	private void promoteAccumulators(Repeated<T> scope) {
```
ashesfall: This method should operate on itself, and return a new Scope.

**Stop ignoring software design.**

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+
+	/**
+	 * Checks whether two array-element references address the same element:
+	 * the same referent variable (by name) at a structurally equal position.
+	 *
+	 * @param a the first reference
+	 * @param b the second reference
+	 * @return true if both references address the same array element
+	 */
+	private static boolean matchesReference(InstanceReference<?, ?> a, InstanceReference<?, ?> b) {
```
ashesfall: This is a duplicate of code that exists elsewhere, and is anyway general and should not be in this class.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	 * assignments addresses one of the promoted elements, and that no assignment
+	 * with an unanalyzable destination touches the array.
+	 *
+	 * @param body all assignments in the loop body
+	 * @param arrayName the array being considered for promotion
+	 * @param nameSet a singleton set containing {@code arrayName}
+	 * @param dests the promoted element references
+	 * @return true if promotion of the array preserves the loop's semantics
+	 */
+	private static boolean isPromotableEverywhere(List<ExpressionAssignment<?>> body, String arrayName,
```
ashesfall: Do not put utilities in this special subclass.

Stop ignoring software design.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	/**
+	 * Checks that every {@link InstanceReference} to the named array within an
+	 * expression tree addresses one of the given element references.
+	 *
+	 * @param expr the expression tree to scan
+	 * @param arrayName the array name to look for
+	 * @param dests the sanctioned element references
+	 * @return true if no reference to the array addresses any other element
+	 */
+	private static boolean referencesMatchDestinations(Expression<?> expr, String arrayName,
```
ashesfall: Do not put utilities in this special subclass.

Stop ignoring software design.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	/**
+	 * Creates one local accumulator reference per promoted element, using the
+	 * loop index name as a uniqueness prefix.
+	 *
+	 * @param dests the promoted element references
+	 * @param prefix the loop index name used to make accumulator names unique
+	 * @param start the starting counter value for accumulator naming
+	 * @return the accumulator references, or null if any element type is unavailable
+	 */
+	private static List<StaticReference<?>> createAccumulators(List<InstanceReference<?, ?>> dests,
```
ashesfall: Do not put utilities in this special subclass.

Stop ignoring software design.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	/**
+	 * Recursively rewrites all assignments in a loop-body scope, replacing each
+	 * reference to a promoted array element with its local accumulator.
+	 *
+	 * @param scope the loop-body scope to rewrite
+	 * @param arrayName the promoted array
+	 * @param dests the promoted element references
+	 * @param accumulators the local accumulator for each promoted element
+	 */
+	private static void rewriteBodyAssignments(Scope<?> scope, String arrayName,
```
ashesfall: Do not put utilities in this special subclass.

Stop ignoring software design.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	 * with their local accumulators in both the destination and the expression.
+	 * Returns the original assignment if nothing was replaced.
+	 *
+	 * @param assignment the assignment to rewrite
+	 * @param arrayName the promoted array
+	 * @param dests the promoted element references
+	 * @param accumulators the local accumulator for each promoted element
+	 * @return a new assignment with replacements, or the original if unchanged
+	 */
+	private static ExpressionAssignment<?> rewriteAssignment(ExpressionAssignment<?> assignment, String arrayName,
```
ashesfall: Do not put utilities in this special subclass.

Stop ignoring software design.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +334,370 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	 * tree with its local accumulator, using position-based child replacement.
+	 * Returns the original expression if no replacements were made.
+	 *
+	 * @param expr the expression to process
+	 * @param arrayName the promoted array
+	 * @param dests the promoted element references
+	 * @param accumulators the local accumulator for each promoted element
+	 * @return the expression with replacements applied, or the original if unchanged
+	 */
+	private static Expression<?> substituteAccumulators(Expression<?> expr, String arrayName,
```
ashesfall: Do not put utilities in this special subclass.

Stop ignoring software design.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -815,6 +1221,7 @@ public Repeated<T> generate(List<Scope<T>> children) {
 		Repeated<T> scope = getMetadata() == null ? new Repeated<>(getName()) : new Repeated<>(getName(), getMetadata());
 		scope.setIndex(getIndex());
 		scope.getChildren().addAll(children);
+		scope.getEpilogue().addAll(getEpilogue());
```
ashesfall: No

base/code/src/main/java/io/almostrealism/scope/ScopeExpressionCollector.java
```diff
@@ -176,6 +177,14 @@ private static void collectFromRepeated(Scope<?> s, List<Expression<?>> roots) {
 			Repeated<?> r = (Repeated<?>) s;
 			if (r.getCondition() != null) roots.add(r.getCondition());
 			if (r.getInterval() != null) roots.add(r.getInterval());
+
+			for (Statement<?> stmt : r.getEpilogue()) {
```
ashesfall: Stop introducing special cases.
Stop ignoring software design.

base/hardware/src/main/java/org/almostrealism/hardware/cl/CLMemory.java
```diff
@@ -57,6 +59,85 @@ public class CLMemory extends RAM {
...
+		if (probeGeneration == generation) return true;
+		probeGeneration = generation;
+		return false;
+	}
+
+	/**
+	 * Discards any host read cache. Called when this buffer is written on the host so a
+	 * subsequent read does not observe stale contents.
+	 */
+	public void invalidateHostCache() {
```
ashesfall: Do not put methods before the constructor, a bunch of this should be private internal functionality of CLMemory which belongs at the bottom anyway.

REALLY bad style choices.

ashesfall: We need a hook to detect this kind of crap too, methods before the constructor should fail the build and we should have a hook to match.

base/hardware/src/main/java/org/almostrealism/hardware/cl/CLMemoryProvider.java
```diff
@@ -440,6 +455,63 @@ public void getMem(CLMemory mem, int sOffset, double out[], int oOffset, int len
...
+	 * {@link CLMemory#currentGeneration() dispatch generation} — so a lone read never triggers a
+	 * full transfer — and is discarded once a kernel dispatch or a host write invalidates it. A
+	 * read that already covers the whole buffer transfers directly and is not retained, so bulk
+	 * transfers do not pay for a cached copy.</p>
+	 *
+	 * @param mem    the memory being read
+	 * @param length the number of elements requested
+	 * @return a whole-buffer snapshot indexed by element, or {@code null} to read directly
+	 */
+	private double[] hostReadCache(CLMemory mem, int length) {
```
ashesfall: Why is this not a method of CLMemory?

base/hardware/src/main/java/org/almostrealism/hardware/cl/CLMemoryProvider.java
```diff
@@ -440,6 +455,63 @@ public void getMem(CLMemory mem, int sOffset, double out[], int oOffset, int len
...
+	}
+
+	/**
+	 * Reads every element of the buffer into a new {@code double[]} with a single transfer.
+	 *
+	 * @param mem   the memory to read
+	 * @param total the number of elements in the buffer
+	 * @return a snapshot of all elements as doubles
+	 */
+	private double[] readWholeBuffer(CLMemory mem, int total) {
```
ashesfall: Why would we have a special method for reading the "whole buffer". This doesn't even do that, it reads a specific number of elements that is passed as an argument!

base/hardware/src/main/java/org/almostrealism/hardware/cl/CLOperator.java
```diff
@@ -218,6 +218,9 @@ public synchronized Semaphore accept(Object[] args, Semaphore dependsOn) {
 
 		long id = totalInvocations++;
 
+		// Invalidate any host read caches: this kernel may write its argument buffers.
+		CLMemory.markDispatch();
```
ashesfall, line 222: Why are you doing this for all CLMemory? Why not just the CLMemory actually touched by this operator? This is insane.

domain/graph/src/main/java/org/almostrealism/CodeFeatures.java
```diff
@@ -420,22 +420,37 @@ default OperationProfileNode profile(String name, Supplier<Runnable> op) {
...
+	 * saved profile.</p>
+	 *
+	 * @param profile the profile node to collect timing and source into; may be {@code null}
 	 * @param op      a supplier producing the runnable to profile
 	 * @return the profile node after the operation completes
 	 */
 	default OperationProfileNode profile(OperationProfileNode profile, Supplier<Runnable> op) {
-		Runnable r;
+		try {
+			Hardware.getLocalHardware().assignProfile(profile);
```
ashesfall: Did you check if this method is used in places that already perform profile assignment before hand and do not expect for it to be cleared afterwards?

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +299,230 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	 * nested {@link Repeated} loop bounds), so any of them causes analysis to fail.
+	 * Method calls, metrics, and statements that are not
+	 * {@link ExpressionAssignment}s fail for the same reason, which makes the
+	 * caller skip accumulator promotion for the whole loop.</p>
+	 *
+	 * @param scope the loop-body scope to scan
+	 * @param assignments the list to add assignments to
+	 * @return true if the body was fully analyzable, false otherwise
+	 */
+	private static boolean collectBodyAssignments(Scope<?> scope, List<ExpressionAssignment<?>> assignments) {
```
ashesfall: This should be an instance method of Scope, and it can return null to indicate the failure ("return false") case.

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +299,230 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	 * their accumulators; if any reference to {@code array} survives, it addresses an
+	 * element that is not being promoted and promotion must be abandoned.
+	 *
+	 * @param body         the assignments in the loop body
+	 * @param array        the array being considered for promotion
+	 * @param elements     the promoted element references
+	 * @param accumulators the accumulator that replaces each promoted element
+	 * @return true if promotion of the array preserves the loop's semantics
+	 */
+	private static boolean referencesOnlyPromotedElements(List<ExpressionAssignment<?>> body, Variable<?, ?> array,
```
ashesfall: Why not an instance method of ExpressionAssignment?

base/code/src/main/java/io/almostrealism/scope/Repeated.java
```diff
@@ -292,6 +299,230 @@ private void hoistLoopInvariantStatements(Repeated<T> scope) {
...
+	 * nor reads an element of an array that the loop writes (whose value would change
+	 * as the loop runs).</p>
+	 *
+	 * @param position     the element position expression to check
+	 * @param loopScalars  names of the loop index and scalars assigned in the loop
+	 * @param storedArrays the arrays written inside the loop
+	 * @param loopIndices  indices from this and nested loops
+	 * @return true if the position addresses the same element on every iteration
+	 */
+	private boolean isPositionInvariant(Expression<?> position, Set<String> loopScalars,
```
ashesfall: Why not an instance method of Expression?

base/hardware/src/main/java/org/almostrealism/hardware/cl/CLMemory.java
```diff
@@ -57,6 +60,22 @@ public class CLMemory extends RAM {
 	/** True once {@code clReleaseMemObject} has been invoked for the underlying handle. */
 	private volatile boolean released;
 
+	/**
+	 * Monotonically increasing counter incremented once per OpenCL kernel dispatch (see
+	 * {@link #markDispatch()}). A host read cache captured at one value of this counter is
+	 * stale once the counter advances, because an intervening kernel may have written the buffer.
+	 */
+	private static final AtomicLong dispatchGeneration = new AtomicLong();
```
ashesfall: Static members belong at the top.

base/hardware/src/main/java/org/almostrealism/hardware/cl/CLMemoryProvider.java
```diff
@@ -440,6 +455,35 @@ public void getMem(CLMemory mem, int sOffset, double out[], int oOffset, int len
...
+	/**
+	 * Reads every element of the buffer into a new {@code double[]} with a single transfer.
+	 * Used by {@link CLMemory#snapshotForRead(int, java.util.function.Supplier)} to capture a
+	 * whole-buffer snapshot that serves repeated per-element reads without a device round-trip
+	 * per element.
+	 *
+	 * @param mem the memory to read
+	 * @return a snapshot of all elements as doubles
+	 */
+	private double[] readWholeBuffer(CLMemory mem) {
```
ashesfall: Why isn't getMem suitable? If getMem and this method must both exist, they need to delegate to a shared method. this is WAY too much code to duplicate.

base/hardware/src/main/java/org/almostrealism/hardware/Hardware.java
```diff
@@ -920,6 +923,7 @@ public void assignProfile(OperationProfile profile) {
 		if (profile == null) {
 			clearProfile();
 		} else {
+			this.profile = profile;
```
ashesfall: The entire purpose of having this method delegate to a bunch of other classes is to avoid a direct reference between a Hardware instance and the profile. Do not do this.

base/hardware/src/main/java/org/almostrealism/hardware/cl/CLMemoryProvider.java
ashesfall: This is now a duplicate of mem.toArray, please remove it

base/code/src/main/java/io/almostrealism/expression/Expression.java
```diff
@@ -585,6 +586,40 @@ public boolean containsStaticReferenceToAny(Set<String> names) {
...
+
+		for (Variable<?, ?> var : getDependencies()) {
+			if (var.getName() != null && loopVariantNames.contains(var.getName())) return false;
+		}
+
+		for (Index idx : getIndices()) {
+			if (idx.getName() != null && loopVariantNames.contains(idx.getName())) return false;
+		}
+
+		if (containsStaticReferenceToAny(loopVariantNames)) return false;
```
ashesfall: In an earlier commit we noticed the containsInstanceReferenceToAny is a method that should not exist, and the same applies to containsStaticReferenceToAny. I will reiterate my comment:

This method name is unnecessarily long, it should be containsReference.

We also should not have so many implementations of containsReference: the containsReference method which accepts a Variable should be preferred (we can accept a Set of Variables too), but if it is prohibitive for users to use Variable then the name option can be introduced but it should work in the way that containsReference does: a default that delegates to a children, and implementations directly in the subclasses that handle the various cases.

This is another serious violation of object oriented principles: you are packing all of the complexity into this parent class rather than embracing the flexibility that is provided by having a type hierarchy in the first place. With methods like this, we might as well just have a single Expression class and let it be in one of many modes.

PR #343, opened 2026-07-09
Use bulk transfers for cross-provider memory copies on the OpenCL path

ashesfall:
Every dispatch of a kernel whose arguments reside on a different memory
provider performs a prepare copy into a temporary buffer on the kernel's
provider and a postprocess copy back (memory replacement). Under
AR_HARDWARE_DRIVER=native,cl those copies moved data through a Java double[]
on every run, and the native side of the transfer called SetDoubleArrayRegion
once per element, yielding roughly 20 MB/s. For a reduction dispatched
repeatedly this made per-run time scale with input buffer size — measured
1.56/2.17/4.56 ms per run at reduction widths 2/16/64 — while the compiled
kernel itself accounted for only ~25 us.

Move these copies to single bulk transfers:

- MemoryProvider gains two capability methods with no-op defaults:
  getHostBuffer(mem, offset, length), returning a direct ByteBuffer view of
  the region, and getMem(mem, sOffset, ByteBuffer, length) for bulk reads.
  Both deal in raw bytes in the provider's native element format, so callers
  confirm getNumberSize() matches before transferring.

- NativeBufferView, a new JNI operation, wraps a range of a native allocation
  in a direct buffer via NewDirectByteBuffer without copying.
  NativeMemoryProvider uses it to implement getHostBuffer, and its
  setMem(..., Memory source, ...) now reads bulk from the source provider
  directly into that view when element sizes match, covering the CL-to-native
  copy-back direction with a single clEnqueueReadBuffer and no intermediate
  arrays.

- CLMemoryProvider.setMem(..., Memory, ...) resolves its long-standing TODO:
  when the source provider exposes a host-buffer view and element sizes
  match, the write is a single clEnqueueWriteBuffer from that view, covering
  the native-to-CL prepare direction. CLMemoryProvider also implements the
  bulk getMem(ByteBuffer) read.

- NativeRead's generated C previously invoked SetDoubleArrayRegion per
  element; it now performs one bulk region copy (FP64) or one conversion pass
  into a temporary plus one region copy (FP32). This benefits every native
  host read, independent of OpenCL.

Under native,cl (uniform FP32, so element sizes always match) the same
reduction now runs at a flat ~1.3 ms per dispatch at every width — the
transfer component is gone. The remaining fixed cost is the replacement
machinery itself (temporary allocation per dispatch, semaphore chaining, and
thread handoffs), which is out of scope here; eliminating it is the subject
of the CLMemory-everywhere/CLJNI direction recorded in
docs/plans/OPENCL_REDUCTION_PERFORMANCE.md.

CrossProviderMemoryCopyTest covers both copy directions with offsets and
sentinel values, round-trips data native -> CL -> native, and skips cleanly
when the OpenCL or JNI backend is unavailable.


base/code/src/main/java/io/almostrealism/code/MemoryProvider.java
```diff
@@ -111,6 +112,45 @@ default double[] toArray(T mem, int offset, int length) {
...
+	 * transfer mechanism return {@code false} and the caller falls back to array-mediated
+	 * reads.</p>
+	 *
+	 * @param mem the source memory region
+	 * @param sOffset the starting position in the source memory, in elements
+	 * @param out the destination buffer, positioned at zero with capacity for the transfer
+	 * @param length the number of elements to read
+	 * @return true if the transfer was performed, false if unsupported
+	 */
+	default boolean getMem(T mem, int sOffset, ByteBuffer out, int length) {
```
ashesfall: We have a Memory implementation that is backed by a ByteBuffer. The correct way to expose a feature like this is to use the EXISTING getMem method, updating implementations of getMem that want to support this with a check if the destination is backed by a ByteBuffer and handling the copy differently in that situation.

base/hardware/src/main/java/org/almostrealism/c/NativeBufferView.java
```diff
@@ -0,0 +1,90 @@
...
+ *
+ * <p>The returned buffer is a view, not a copy: it remains valid only while the
+ * underlying allocation is live. Callers must keep the owning {@link NativeMemory}
+ * strongly reachable for as long as the buffer is in use.</p>
+ *
+ * @see NativeMemory
+ * @see NativeMemoryProvider
+ * @see BaseNative
+ */
+public class NativeBufferView extends BaseNative {
```
ashesfall, line 47: This duplicates functionality that our NIO memory provider already supports. If you actually need this, then the NIO memory provider will need to be updated to use it.

Stop engaging in rampant code duplication.

.github/workflows/analysis.yaml
```diff
@@ -813,6 +813,12 @@ jobs:
        needs.changes.outputs.compute_changed == 'true' ||
        needs.changes.outputs.base_changed == 'true')
     runs-on: [self-hosted, linux, ar-ci]
+    # Native memory allocates as direct ByteBuffers (see test-media). Because direct
```
ashesfall: This comment is way too verbose; we do not need to narrate everything everywhere.

.github/workflows/analysis.yaml
```diff
@@ -971,6 +977,23 @@ jobs:
        needs.changes.outputs.compute_changed == 'true' ||
        needs.changes.outputs.base_changed == 'true')
     runs-on: [self-hosted, linux, ar-ci]
+    # Native memory allocates as direct ByteBuffers, which count against the JVM's
```
ashesfall: This comment is way too verbose; we do not need to narrate everything everywhere.

.github/workflows/analysis.yaml
```diff
@@ -1316,6 +1371,21 @@ jobs:
        needs.changes.outputs.compute_changed == 'true' ||
        needs.changes.outputs.base_changed == 'true')
     runs-on: [self-hosted, macos, ar-ci]
+    # Same as test-media: native,cl runs the media tests on direct-ByteBuffer native RAM
```
ashesfall: This comment is way too verbose; we do not need to narrate everything everywhere.

.github/workflows/analysis.yaml
```diff
@@ -1316,6 +1371,21 @@ jobs:
...
+    # (no Metal). Cap direct memory at 8g (below the runner's RAM, so a spike is a catchable
+    # limit rather than an OOM kill) and split the tests across three groups — the CL backend's
+    # direct-ByteBuffer reservations exceed the two-group budget, so it needs one more split than
+    # test-media-mac (still two groups under Metal). See test-media.
+    env:
+      _JAVA_OPTIONS: -XX:MaxDirectMemorySize=8g
+    strategy:
+      fail-fast: false
+      # Cap concurrency at 2 even with 3 groups so this job does not monopolize the
```
ashesfall: This comment is way too verbose; we do not need to narrate everything everywhere.

base/hardware/src/main/java/org/almostrealism/nio/Free.java
```diff
@@ -55,7 +55,7 @@
  *
  * @see Malloc
  * @see BaseNative
- * @see NativeMemoryProvider
+ * @see org.almostrealism.nio.NativeMemoryProvider
  */
 public class Free extends BaseNative {
```
ashesfall, line 61: This class should move to the nio package along with the provider.

base/hardware/src/main/java/org/almostrealism/nio/Malloc.java
```diff
@@ -43,15 +43,15 @@
...
  *
  * <p>The native {@code malloc()} function is thread-safe. Multiple threads
  * can allocate memory concurrently.</p>
  *
  * @see Free
- * @see NativeMemoryProvider
+ * @see org.almostrealism.nio.NativeMemoryProvider
  * @see BaseNative
  */
 public class Malloc extends BaseNative {
```
ashesfall, line 58: This class should move to the nio package along with the provider.

domain/graph/src/main/java/org/almostrealism/graph/AdjustableDelayCell.java
```diff
@@ -232,7 +232,7 @@ public void reset() {
 	 *
 	 * <p>The {@link AcceleratedTimeSeries} buffer is the largest allocation
 	 * (10M entries by default). Destroying it returns the native memory to
-	 * the {@link org.almostrealism.c.NativeMemoryProvider} immediately rather
+	 * the {@link org.almostrealism.nio.NativeMemoryProvider} immediately rather
```
ashesfall: This comment is just wrong - we don't know what MemoryProvider will be used, and you should not mention any specific MemoryProvider type in the comment. This whole sentence is stupid anyway, everyone knows what destroy means.

base/hardware/src/main/java/org/almostrealism/nio/NativeRead.java
```diff
@@ -63,7 +63,7 @@
  *
  * @see NativeWrite
  * @see NativeMemory
- * @see NativeMemoryProvider
+ * @see org.almostrealism.nio.NativeMemoryProvider
  * @see BaseNative
  */
 public class NativeRead extends BaseNative {
```
ashesfall, line 70: This should move to the nio package with the others.

base/hardware/src/main/java/org/almostrealism/nio/NativeWrite.java
```diff
@@ -56,7 +56,7 @@
  *
  * @see NativeRead
  * @see NativeMemory
- * @see NativeMemoryProvider
+ * @see org.almostrealism.nio.NativeMemoryProvider
  * @see BaseNative
  */
 public class NativeWrite extends BaseNative {
```
ashesfall, line 63: This should move to the nio package with the others.

base/hardware/src/main/java/org/almostrealism/c/package-info.java
```diff
@@ -18,7 +18,7 @@
  * C language code generation and native memory management infrastructure.
  *
  * <p>This package provides C language operations for generating kernel source code,
- * native memory allocation via JNI ({@link org.almostrealism.c.NativeMemoryProvider}),
+ * native memory allocation via JNI ({@link org.almostrealism.nio.NativeMemoryProvider}),
```
ashesfall: This comment makes no sense if the package changes, because its describing a class in a different package.

base/hardware/src/main/java/org/almostrealism/hardware/mem/ByteBufferMemory.java
```diff
@@ -0,0 +1,50 @@
...
+ * type, and fall back to array-mediated transfers for any other {@link RAM}.</p>
+ *
+ * <p>Declaring the capability here — in the shared {@code org.almostrealism.hardware.mem}
+ * package alongside {@link RAM} — rather than on a concrete implementation keeps consumers
+ * decoupled from the module that provides the backing memory, so an implementation can later
+ * move to its own module without every consumer depending on it.</p>
+ *
+ * @see RAM
+ */
+public abstract class ByteBufferMemory extends RAM {
```
ashesfall: This is a terrible name, it describes something superficial about the type ("can be viewed as ByteBuffer") rather than something about the actual kind of thing this is. The name DirectMemory is better (since it fits the pattern of how the JVM refers to this kind of thing, as in the method allocateDirect).

base/hardware/src/main/java/org/almostrealism/hardware/mem/ByteBufferMemory.java
```diff
@@ -0,0 +1,50 @@
...
+	 * Returns a direct {@link ByteBuffer} over this memory's contents.
+	 *
+	 * <p>The buffer shares storage with the memory — it is a view, not a copy — and holds the
+	 * raw bytes in the backing element format, with no conversion. It remains valid only while
+	 * this memory is allocated, so a caller must keep this instance strongly reachable for as
+	 * long as the buffer is in use.</p>
+	 *
+	 * @return a direct buffer over this memory's contents
+	 */
+	public abstract ByteBuffer getByteBuffer();
```
ashesfall: This is not a getter, in the conventional sense (the buffer may not exist at the time the method is called). The method name asByteBuffer is better.

base/hardware/src/main/java/org/almostrealism/hardware/Hardware.java
```diff
@@ -585,7 +585,7 @@ public final class Hardware implements ConsoleFeatures {
 	/** Location strategy for CL memory allocation (DEVICE, HEAP, HOST, or DELEGATE). */
 	private Location location;
 	/** Optional NIO-based memory provider for Metal/NIO shared memory mode; null if not enabled. */
-	private NativeBufferMemoryProvider nioMemory;
+	private NativeMemoryProvider nioMemory;
```
ashesfall, line 588: This is potentially a problem, for a range of reasons:
(1) I'm not sure that we are adequately preventing the case where we end up with two NativeMemoryProviders, one here and one in the actual native DataContext
(2) When NIO memory is active, the NativeMemoryProvider *has to be created in a way that can be shared* and we aren't enforcing that anywhere

It is insane to imagine that you can just perform a substitution here and everything is fine. Study what is ACTUALLY going on in the case where this is active, and determine what makes sense. This is NOT a problem you can solve in your favored "search and replace" manner.

base/hardware/src/main/java/org/almostrealism/c/NativeMemory.java
```diff
@@ -38,17 +42,22 @@ public class NativeMemory extends RAM {
...
 	/**
 	 * Creates a new NativeMemory wrapping a native memory pointer.
 	 *
 	 * @param provider      the memory provider that allocated this buffer
 	 * @param nativePointer the native pointer to the allocated memory
 	 * @param size          the size of the allocation in bytes
+	 * @param byteBuffer    a non-owning direct buffer view over the whole allocation
 	 */
-	public NativeMemory(MemoryProvider provider, long nativePointer, long size) {
+	public NativeMemory(MemoryProvider provider, long nativePointer, long size, ByteBuffer byteBuffer) {
```
ashesfall: I don't see how it could be valid to accept a native point and a ByteBuffer. Either the native pointer tells us the location, and NativeMemory should be able to derive a ByteBuffer from it; or the ByteBuffer is the source of truth, and we should not allow a native pointer. Is there some reason it could ever be valid to define a constructor that accepts both? What happens when I create this with a native pointer that points to one memory region, and a ByteBuffer that points somewhere else? You surely couldn't have left that case to just be discovered at runtime, so there must be an exception somewhere that I'm not seeing. Where?

base/hardware/src/main/java/org/almostrealism/nio/NIO.java
```diff
@@ -16,41 +16,74 @@
...
  *
- * <p>Loads the platform-specific native library ({@code libNIO.dylib} on macOS, {@code libNIO.so} on Linux)
- * from the classpath and exposes low-level memory-mapped I/O operations for use by
- * {@link NativeBufferMemoryProvider}.</p>
+ * <p>The underlying operations are compiled at runtime by {@link NativeCompiler} — the same way
+ * {@link org.almostrealism.c.Malloc} and the other native operations are — rather than loaded from a
+ * prebuilt platform library. As a result they need no {@code libNIO} artifact and work on any
+ * platform the native backend supports. The operations are compiled lazily on first use.</p>
  */
 public class NIO {
```
ashesfall: Now that we are using dedicated classes for each operation, this class no longer makes sense - we can just go directly to the operations that matter and use them.

base/hardware/src/main/java/org/almostrealism/nio/NativeMemoryProvider.java
```diff
@@ -0,0 +1,442 @@
...
+	 * Whether native (JNI) execution allocates NIO direct buffers ({@link NativeBuffer}) by default
+	 * (the default) rather than JNI {@link Malloc} blocks ({@link NativeMemory}). Both are
+	 * host-accessible {@link org.almostrealism.hardware.mem.ByteBufferMemory} and both work on every
+	 * platform the native backend supports (the direct-buffer pointer and the calloc ByteBuffer view
+	 * are alike produced by runtime-compiled code — no prebuilt library). The difference is only how
+	 * memory is allocated and tracked: direct buffers participate in the JVM's
+	 * {@code -XX:MaxDirectMemorySize} accounting, calloc blocks do not. Controlled by
+	 * {@code AR_HARDWARE_DIRECT_MEMORY}.
+	 */
+	public static boolean enableDirectAllocation =
```
ashesfall: This should be a parameter of the constructor, and the check of SystemUtils should live in Hardware and be threaded through to here.

I also very much suspect that AR_HARDWARE_DIRECT_MEMORY is redundant with respect to Hardwares "nio memory" option, but perhaps there is a difference - if so, the settings need to be renamed to make it more clear *what* that difference is.

PR #347, opened 2026-07-12
Prevent direct MemoryData assignment from Java (Phase 0)

ashesfall:


compute/algebra/src/main/java/org/almostrealism/algebra/Pair.java
```diff
@@ -131,7 +132,7 @@ public Pair(MemoryData delegate, int delegateOffset) {
 	 */
 	public Pair(double x, double y) {
 		this();
-		this.setMem(x, y);
+		CollectionFeatures.getInstance().a(CollectionFeatures.getInstance().cp(this), CollectionFeatures.getInstance().c(x, y)).get().run();
```
ashesfall: This is not necessary, direct subclasses of PackedCollection can be excluded from the requirement to use producers this way.

compute/algebra/src/main/java/org/almostrealism/algebra/Tensor.java
```diff
@@ -254,7 +255,7 @@ public PackedCollection pack() {
 		PackedCollection c = new PackedCollection(shape);
 
 		AtomicInteger index = new AtomicInteger();
-		shape.stream().forEach(pos -> c.setMem(index.getAndIncrement(), ((Number) get(pos)).doubleValue()));
```
ashesfall, line 257: This should be whitelisted for now.

compute/algebra/src/main/java/org/almostrealism/collect/CollectionFeatures.java
```diff
@@ -484,7 +484,7 @@ default CollectionProducerComputation c(TraversalPolicy shape,
 				PackedCollection col = c.evaluate(args);
 				PackedCollection idx = i.evaluate(args);
 				PackedCollection dest = out.evaluate(args);
-				dest.setMem(col.toDouble((int) idx.toDouble(0)));
+				a(cp(dest), c(col.toDouble((int) idx.toDouble(0)))).get().run();
```
ashesfall: Putting a kernel in the short-circuit defeats the purpose of the short-circuit mechanism! This should use setFrom

compute/algebra/src/main/java/org/almostrealism/collect/SlicingFeatures.java
```diff
@@ -693,13 +694,13 @@ default CollectionProducer cumulativeProduct(Producer<PackedCollection> input, b
 			int offset = 0;
 
 			if (pad) {
-				result.setMem(0, r);
+				CollectionFeatures.getInstance().a(CollectionFeatures.getInstance().cp(result.range(new TraversalPolicy(1), 0)), CollectionFeatures.getInstance().c(r)).get().run();
```
ashesfall: This is AWFUL. We should actually create a Computation here. This method *never should have used Java AT ALL* and now you are embedding a kernel inside of a function that was supposed to be a kernel? This is completely insane.

PR #350, opened 2026-07-13
Prevent direct MemoryData assignment from Java (Phase 2, 3, 4)

ashesfall:


domain/heredity/src/main/java/org/almostrealism/heredity/ScaleFactor.java
```diff
@@ -106,8 +98,7 @@ public Producer<PackedCollection> getResultant(Producer<PackedCollection> value)
 	 * @param s the new scale value
 	 */
 	public void setScaleValue(double s) {
```
ashesfall: We should treat uses of this method with non-literal double values, as violations.

studio/music/src/main/java/org/almostrealism/music/pattern/PatternLayerManager.java
```diff
@@ -615,8 +616,12 @@ public void layer(Gene<PackedCollection> gene) {
 	protected void layer(ParameterSet params) {
 		Gene<PackedCollection> automationGene = envelopeAutomationChromosome.valueAt(depth());
 		PackedCollection automationParams =
-				PackedCollection.factory().apply(AUTOMATION_GENE_LENGTH).fill(pos ->
-						automationGene.valueAt(pos[0]).getResultant(null).evaluate().toDouble());
+				PackedCollection.factory().apply(AUTOMATION_GENE_LENGTH);
+		concat(shape(AUTOMATION_GENE_LENGTH),
```
ashesfall: Is there a test to confirm that the new code here produces the same result as the old? There is some risk here, potentially.

compute/algebra/src/main/java/org/almostrealism/algebra/computations/LoopedWeightedSumComputation.java
```diff
@@ -148,6 +149,11 @@ public LoopedWeightedSumComputation(String name,
 		this.weightShape = weightShape;
 		this.inputIndexer = inputIndexer;
 		this.weightIndexer = weightIndexer;
+
+		// The signature recorded during superclass construction was computed
+		// before the indexers above were assigned, so it must be refreshed
+		// now that the state is complete
+		setMetadata(getMetadata().withSignature(signature()));
```
ashesfall: The normal pattern here is to call init() over again, and you should not leave such a detailed inline comment for such a minor thing.

compute/algebra/src/main/java/org/almostrealism/algebra/computations/LoopedWeightedSumComputation.java
```diff
@@ -248,4 +254,44 @@ public LoopedWeightedSumComputation generate(List<Process<?, ?>> children) {
...
+			String indexers = inputIndexer.index(output, outer, inner)
+					.getExpression(Expression.defaultLanguage()) + ";" +
+					weightIndexer.index(output, outer, inner)
+							.getExpression(Expression.defaultLanguage());
+
+			return signature + "{innerCount:" + innerCount +
+					",input:" + inputShape.toStringDetail() +
+					",weight:" + weightShape.toStringDetail() +
+					",indexers:" + indexers + "}";
+		} catch (Exception e) {
```
ashesfall: Catching an exception as a solution to a standard control flow concern is a big mistake. What is it that you need to know, which you can know by asking? Expression (or Index) should give you the answer, not incur the expense of exception handling.

compute/algebra/src/main/java/org/almostrealism/collect/computations/AggregatedProducerComputation.java
```diff
@@ -272,19 +278,27 @@ public boolean isReplaceLoop() {
 	 */
 	public AggregatedProducerComputation setReplaceLoop(boolean replaceLoop) {
 		this.replaceLoop = replaceLoop;
+
+		// Loop replacement changes the generated kernel, so the signature
+		// recorded when this computation was constructed must be refreshed
+		setMetadata(getMetadata().withSignature(signature()));
```
ashesfall: Once we're doing things like "the signature depends on this", the field cannot be mutable this way.

What if we put this in a Map by signature? and you just change it?

Come on... THINK

compute/algebra/src/main/java/org/almostrealism/collect/computations/AggregatedProducerComputation.java
```diff
@@ -499,6 +513,54 @@ public AggregatedProducerComputation generate(List<Process<?, ?>> children) {
...
+	protected String expressionSignature() {
+		try {
+			Expression left = new StaticReference(Double.class, "signatureLeft");
+			Expression right = new StaticReference(Double.class, "signatureRight");
+			String combined = expression.apply(left, right)
+					.getExpression(Expression.defaultLanguage());
+			String start = initial.apply(new TraversableExpression[0], e(0))
+					.getExpression(Expression.defaultLanguage());
+			return start + ";" + combined;
+		} catch (Exception e) {
```
ashesfall: Do not use exceptions for this, ESPECIALLY not Exceptions with no type.

compute/algebra/src/main/java/org/almostrealism/algebra/computations/WeightedSumComputation.java
```diff
@@ -135,6 +135,52 @@ public WeightedSumComputation(TraversalPolicy resultShape,
 				weightGroupShape.getDimensions() != weightShape.getDimensions()) {
 			throw new IllegalArgumentException();
 		}
+
+		// The signature recorded during superclass construction was computed
+		// before the traversal policies above were assigned, so it must be
+		// refreshed now that the state is complete
+		setMetadata(getMetadata().withSignature(signature()));
```
ashesfall: The standard for this is to call init() over again.

PR #353, opened 2026-07-15
Prevent direct MemoryData assignment from Java (Phase 8) 

ashesfall:


engine/ml/src/main/java/org/almostrealism/ml/AttentionFeatures.java
```diff
@@ -172,6 +172,32 @@
...
+	 * index table in memory and move it from the host.</p>
+	 *
+	 * @param name         the name of the resulting layer
+	 * @param inputShape   the shape of the layer's input
+	 * @param outputShape  the shape of the layer's output
+	 * @param indices      producer of one input index per output element
+	 * @param requirements optional compute requirements
+	 * @return a {@link CellularLayer} gathering input elements by index
+	 */
+	default CellularLayer gather(String name, TraversalPolicy inputShape, TraversalPolicy outputShape,
```
ashesfall, line 191: This does not belong in AttentionFeatures, it is clearly more general than that.

PR #355, opened 2026-07-16
Prevent direct MemoryData assignment from Java (Phase 9) 

ashesfall:


ashesfall:
We have a passing CI pipeline for this (https://github.com/almostrealism/common/actions/runs/29551780496), but the code itself still needs to be reviewed.

engine/audio/src/main/java/org/almostrealism/audio/computations/ADSREnvelopeComputation.java
```diff
@@ -0,0 +1,170 @@
...
+				position.add(e(dt).divide(time)), e(1.0));
+	}
+
+	/**
+	 * Clamps an expression to a maximum of one, the device equivalent of {@code min(1, x)}.
+	 *
+	 * @param x the expression to clamp
+	 * @return {@code min(1, x)}
+	 */
+	private Expression clampToOne(Expression x) {
```
ashesfall: Why not use the Expression class Min?

If Min is broken, we need to fix it rather than try and circumvent it.

engine/audio/src/main/java/org/almostrealism/audio/filter/BiquadFilterData.java
```diff
@@ -101,4 +104,249 @@ default void resetState() {
...
+	 * @param center     center frequency in Hz, as a {@code (1)} producer
+	 * @param q          Q factor, as a {@code (1)} producer
+	 * @param sampleRate sample rate in Hz
+	 * @return a {@code (5)} producer of {@code [b0, b1, b2, a1, a2]}
+	 */
+	default CollectionProducer bandPassCoefficients(
+			Producer<PackedCollection> center, Producer<PackedCollection> q, int sampleRate) {
+		CollectionProducer cosW0 = cos(omega(center, sampleRate));
+		CollectionProducer alpha = alpha(center, q, sampleRate);
+		return simpleCoefficients(alpha, c(0.0), alpha.multiply(-1.0), cosW0, alpha);
```
ashesfall: Please use alpha.minus(), rather than multiply by negative 1.

engine/audio/src/main/java/org/almostrealism/audio/synth/AudioSynthesizer.java
```diff
@@ -329,6 +348,7 @@ public void setHighPassFilter(double cutoffHz, double resonance) {
 	 */
 	public void setBandPassFilter(double centerHz, double q) {
 		this.filterBaseCutoff = centerHz;
+		this.filterBaseCutoffData.fill(centerHz);
```
ashesfall, line 343: Why do we need to maintain two copies of this information? Can't the PackedCollection be the source of truth, and just make double fields like filterBaseCutoff unnecessary?

ashesfall: Agreed — there's no reason to keep both. I removed the `filterBaseCutoff` and `filterEnvelopeAmount` `double` fields; the one-element `PackedCollection` buffers (`filterBaseCutoffData`, `filterEnvelopeAmountData`) are now the single source of truth. The setters `fill(...)` them directly, `getFilterEnvelopeAmount()` and the build-time `amount != 0` guard in `tick()` read them via `toDouble(0)` (both control-boundary reads, not per-sample), and the per-tick device op reads them via `cp(...)`. The constructor initializes the buffers to the previous defaults (5000 Hz base, 0 amount). No behavior change — `AudioSynthesizerFilterModulationTest` still passes.

engine/audio/src/main/java/org/almostrealism/audio/filter/BiquadFilterData.java
ashesfall: Done — changed `alpha.multiply(-1.0)` to `alpha.minus()` here (band-pass `b2`), and applied the same to the identical `onePlusCos.multiply(-1.0)` in `highPassCoefficients` (`b1`), since it's the same negate-by-multiply pattern. I left the `.multiply(-2.0)` terms as-is since those are a scale, not a negation. `BiquadCoefficientParityTest` still matches the host RBJ reference for both the high-pass and band-pass responses.

engine/audio/src/main/java/org/almostrealism/audio/computations/ADSREnvelopeComputation.java
ashesfall: You're right — `Min` isn't broken; I reached for a conditional out of expedience without checking for it. `Min.of` folds constant operands, emits the language `min()` via `lang.min(...)`, and evaluates with `Math.min`, which is exactly what's wanted here. I removed the `clampToOne` helper (which was `conditional(x > 1 ? 1 : x)`) and replaced it with `Min.of(x, e(1.0))` inline at the three level expressions. `ADSREnvelopeParityTest` still passes and its negative control still fails, so behavior is unchanged.

PR #356, opened 2026-07-17
Memory use improvements for AudioScene

ashesfall:


base/hardware/src/main/java/org/almostrealism/hardware/mem/Heap.java
```diff
@@ -1117,4 +1162,26 @@ public static void addPendingKernel(Semaphore sem) {
...
+	 * between genome evaluations in an optimizer whose entire run shares one
+	 * {@link #wrap(Callable)} — to release the per-invocation kernel argument
+	 * destinations that {@link #addCreatedMemory(MemoryData)} accumulates. Without
+	 * a boundary flush, those buffers are held until the heap itself is destroyed,
+	 * which can exhaust the memory provider's capacity over many rounds
+	 * ({@code HardwareException} "Memory Max Reached").</p>
+	 *
+	 * @see HeapStage#flushCreatedMemory()
+	 */
+	public static void flushCreatedMemory() {
```
ashesfall: This is a very bad idea.

studio/compose/src/main/java/org/almostrealism/studio/optimize/AudioScenePopulation.java
```diff
@@ -211,13 +212,27 @@ private void enableGenome(Genome newGenome) {
...
+	 */
 	@Override
 	public void disableGenome() {
 		this.currentGenome = null;
 
 		if (temporal != null) {
 			temporal.reset();
 		}
+
+		Heap.flushCreatedMemory();
```
ashesfall: Even if we allowed a method like this to exist, calling it here deep inside the stack is a recipe for callers to get unexpected behavior.

studio/compose/src/main/java/org/almostrealism/studio/AudioScene.java
```diff
@@ -605,12 +605,22 @@ public void loadPatterns(String patternsFile) throws IOException {
...
+	 * device memory accumulates across an optimizer's genome loop until the memory
+	 * provider's cap is reached.</p>
+	 *
 	 * @param genome the new genome whose parameters will be assigned
 	 */
 	public void assignGenome(ProjectedGenome genome) {
 		this.genome.assignTo(genome.getParameters());
 		this.progression.refreshParameters();
 		this.patterns.refreshParameters();
+		PatternLayerManager.invalidateCaches();
```
ashesfall, line 623: This is not a good place to take an action like this.

PR #360, opened 2026-07-20
Adjust audio tests for CI pipeline

ashesfall:


base/io/src/main/java/org/almostrealism/io/SystemUtils.java
```diff
@@ -89,6 +96,81 @@ public static boolean isMacOS() {
 		return System.getProperty("os.name", "").contains("Mac OS X");
 	}
 
+	/** Milliseconds the sustained probe write may block before the audio output device is treated as unusable. */
+	private static final long AUDIO_PROBE_TIMEOUT_MS = 2000;
```
ashesfall, line 100: Fields belong at the top.

PR #368, opened 2026-07-30
Prevent direct MemoryData assignment from Java (Phase 11, 12, 13) 

ashesfall:


studio/music/src/main/java/org/almostrealism/music/sequence/GridSequencer.java
```diff
@@ -167,8 +167,8 @@ public WaveDataProviderList create(Producer<PackedCollection> x, Producer<Packed
 				.grid(bpm.l(getStepSize() * getStepCount()), getStepCount(),
 						(IntFunction<Producer<PackedCollection>>) i -> () -> args -> {
 							ParameterSet params = new ParameterSet(evX.evaluate().toDouble(0), evY.evaluate().toDouble(0), evZ.evaluate().toDouble(0));
-							PackedCollection s = new PackedCollection(1);
-							s.setMem(sequence.apply(i).apply(params));
+							double value = sequence.apply(i).apply(params);
+							PackedCollection s = new PackedCollection(1).fill(value);
```
ashesfall: This can all go on one line, and should favor pack over new + fill:
return pack(sequence.apply(i).apply(params));

studio/music/src/test/java/org/almostrealism/music/pattern/BatchedNoteGatherTest.java
```diff
@@ -77,21 +77,14 @@ public class BatchedNoteGatherTest extends TestSuiteBase implements TemporalFeat
 
 	/** Creates a single-element {@link PackedCollection} containing the given value. */
 	private PackedCollection single(double value) {
```
ashesfall: This method is identical to pack(), and should be removed.

studio/spatial/src/main/java/org/almostrealism/spatial/EditableSpatialWaveDetails.java
```diff
@@ -124,7 +125,8 @@ public void applyValues(List<SpatialValue<?>> values, TemporalSpatialContext con
 
 			int dataIndex = frameIndex * frequencyBins + binIndex;
 			double existing = freqData.toDouble(dataIndex);
-			freqData.setMem(dataIndex, Math.max(existing, magnitude));
+			double updated = Math.max(existing, magnitude);
+			freqData.range(new TraversalPolicy(1), dataIndex).fill(updated);
```
ashesfall: Using range to make the fill method into an alternative to setMem(index, value) is cheating.

We have an existing CollectionProducer Computation which selects the value at a particular index and there are some pre-existing Assignment unit tests that illustrate how to use this pattern to create a Computation which assigns a value to an arbitrary position.

In short, this is not an excusable whitelist case.

studio/spatial/src/test/java/org/almostrealism/spatial/test/SampleBrushVisualTest.java
```diff
@@ -147,7 +148,7 @@ private void paintDots(PackedCollection freqData, int bin, int spacing, double m
 					int b = bin + db;
 					if (b < 0 || b >= FREQ_BINS) continue;
 					int idx = (frame + df) * FREQ_BINS + b;
-					freqData.setMem(idx, magnitude);
+					freqData.range(new TraversalPolicy(1), idx).fill(magnitude);
```
ashesfall: Using range to make fill() into an alternative to setMem(index, value) is cheating and now that you seem to be using it quite widely we need to find a way to block it more specifically.

This whole process can be converted to a Computation; there is no excusable whitelist case here.

PR #371, opened 2026-07-31
Prevent direct MemoryData assignment from Java (Phase 15, 16)

ashesfall:


studio/experiments/src/test/java/org/almostrealism/studio/generate/test/DrawingGenerationEndToEndTest.java
```diff
@@ -166,7 +168,7 @@ public void generateAudioFromDrawing() throws Exception {
 
 			PackedCollection position = new PackedCollection(AudioModel.DIM);
 			for (int i = 0; i < AudioModel.DIM; i++) {
-				position.setMem(i, Math.random());
```
ashesfall, line 169: This change should have caused media tests to run again, but the module must not be included (are we running the tests in this module during CI at all?)

PR #372, opened 2026-07-31
MemoryProviders for ONNX and Protobuf

ashesfall:


base/hardware/src/main/java/org/almostrealism/hardware/mem/ProvidedBytes.java
```diff
@@ -0,0 +1,55 @@
...
+ * <p>This is the delegation target for collections whose backing store is a
+ * {@linkplain SourceMemoryProvider source provider} — a serialized message,
+ * an NIO buffer, a mapped file. Constructing a collection over an instance of
+ * this type touches no device memory at all; migration and per-dispatch
+ * aggregation are handled by the framework exactly as for any other root
+ * reservation.</p>
+ *
+ * @see SourceMemoryProvider
+ */
+public class ProvidedBytes extends MemoryDataAdapter {
```
ashesfall: I don't like the choice of name here, and I also think that we probably should be sharing a parent here between this class and the regular Bytes MemoryData implementation.

If the difference between this class and the existing Bytes class is that we don't do any operations on creation, I think this hardly requires a dedicated class anyway but if it does the more rational name would be something like DeferredBytes.

base/hardware/src/main/java/org/almostrealism/hardware/mem/FloatBufferMemoryProvider.java
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Read-only {@link SourceMemoryProvider} over java.nio {@link FloatBuffer}s,
+ * so results handed over by external runtimes act as their own kind of device:
+ * reads are served from the buffer, and data reaches the compute device only
+ * when a kernel first requires it. Results that only the host ever reads
+ * never occupy device memory at all.
+ *
+ * @see FloatBufferMemory
+ */
+public class FloatBufferMemoryProvider extends SourceMemoryProvider {
```
ashesfall: I understand that this does something slightly different, but it is a real problem that we are creating a new class that implements a *subset* of the functionality that NativeMemoryProvider provides.

At the very least, we need to have any class like this share a parent with NativeMemoryProvider so that overlapping functionality can be accounted for. But more importantly, this FloatBufferMemoryProvider::allocate repeats something that NativeMemoryProvider already does in a far superior way. NativeMemoryProvider: (1) allocates a ByteBuffer and can then provide FloatBuffer or DoubleBuffer as views into the ByteBuffer, preventing us from having to have a new MemoryProvider implementation for every single size number we might deal with in the future (2) supports a JVM-integrated GC solution for the allocations it makes (3) is capable of performing allocation in direct mode (via JVM system library) or itself (via our own system library), each of which have different trade-offs.

NativeMemoryProvider may not be the perfect solution for this use case (if the global Precision is not 32, it may not use FloatBuffer / it doesn't give you control over this separate from the global setting; etc), but that is no reason to ignore it completely. We can make it customizable in more ways (eg, support varying precision values) or in the worst case, create a shared parent for just the overlapping function. But an entirely new class to do a tiny subset of something we already do elsewhere is not acceptable.

ashesfall: FloatBufferMemoryProvider and FloatBufferMemory are deleted. pack(OnnxTensor) now follows the ByteBuffer standard: it allocates staging from Hardware.getNativeBufferMemoryProvider(), writes the tensor's values through a FloatBuffer/DoubleBuffer view of the allocation's ByteBuffer (chosen by the provider's number size), and roots the collection over that memory with Bytes.of. No new provider class, no host arrays, and the staging allocation gets NativeMemoryProvider's GC tracking and direct/calloc modes for free. One contract change worth your eye: getNativeBufferMemoryProvider() now lazily creates the shared bridge when AR_HARDWARE_NIO_MEMORY did not, since ingest needs a staging provider unconditionally.

base/hardware/src/main/java/org/almostrealism/hardware/mem/ProvidedBytes.java
ashesfall: ProvidedBytes is deleted -- it turned out to not merely deserve no dedicated class, but to already exist: Bytes has a private Bytes(Memory, int) constructor with a public Bytes.of(Memory, int) factory. decodeDeferred and the ONNX ingest path now use Bytes.of directly, so no new type was needed at all.

base/hardware/src/main/java/org/almostrealism/hardware/mem/SourceMemoryProvider.java
```diff
@@ -0,0 +1,93 @@
...
+ * aggregate per invocation, so host-only data never permanently occupies
+ * device memory). Migration is one-way: writes into source-backed memory are
+ * rejected rather than silently lost, and nothing ever writes back to the
+ * external source.</p>
+ *
+ * <p>Subclasses supply the format-specific {@code Memory} handles through
+ * their own factory methods; {@link #allocate(int)} is unsupported because a
+ * source provider cannot create empty regions.</p>
+ */
+public abstract class SourceMemoryProvider implements MemoryProvider<Memory> {
```
ashesfall: It seems to me that all of the methods here except for getNumberSize should actually be defaults in the MemoryProvider interface.

After this change, I believe it becomes hare to defend the existence of this class: if it was not abstract, we could call it the NoOpMemoryProvider (and take the number size as an argument to the constructor so anyone could use it as an instance); but as it is this essentially would provide a default value for number size which, to be safe against misunderstandings, really should belong with the subclass.

engine/ml/src/main/java/org/almostrealism/persist/assets/CollectionEncoder.java
```diff
@@ -101,6 +102,32 @@ public static PackedCollection decode(Collections.CollectionData data) {
...
+	 * message on demand and the framework migrates the data to the compute
+	 * device only when a kernel first requires it. Data that is only ever read on the
+	 * host never reaches a device at all. Migration is one-way — the returned
+	 * collection must be treated as read-only until it has migrated.</p>
+	 *
+	 * @param data The protobuf message to serve as the backing store
+	 * @return A collection with the encoded shape reading from the message,
+	 *         or {@code null} if the encoded shape has zero dimensions
+	 */
+	public static PackedCollection decodeDeferred(Collections.CollectionData data) {
```
ashesfall: The preferred pattern here should be a decode method, with a parameter called "materialize" (the pre-existing decode method should still exist, but delegate to the new one with materialize set to false).

When materialize is true, a PackedCollection should be allocated the normal way (via new PackedCollection) and then the "deferred" collection should be copied into the new one using PackedCollection::setFrom.

flowtree/graphpersist/src/test/java/io/almostrealism/db/test/DatabaseTest.java
```diff
@@ -16,7 +16,7 @@ public class DatabaseTest extends TestSuiteBase {
 	 */
 	private static PackedCollection scalarValue(double value) {
```
ashesfall: This is a clone of the pack() method, and it should not exist (if pack() is not available, PackedCollection::of should be used - but a method like this one should never exist)

ashesfall: scalarValue is deleted; the call sites use pack(1.0) through pack(6.0) directly (available on the test through TestSuiteBase's feature surface).

engine/ml/src/main/java/org/almostrealism/persist/assets/CollectionEncoder.java
ashesfall: Done: decodeDeferred is replaced by decode(data, materialize), and the pre-existing decode(data) delegates to it with materialize false -- so deferred is now the default for plain decode. When materialize is true, a collection is allocated the normal way and the deferred one is copied in with setFrom, after which the message-backed root is destroyed. StateDictionary passes materialize = !enableDeferredWeights, and the provider test now covers the materialized path too (values match, ordinary writable storage).

base/hardware/src/main/java/org/almostrealism/hardware/mem/SourceMemoryProvider.java
ashesfall: Agreed and done: allocate(int), both setMem overloads, and destroy() are now defaults on the MemoryProvider interface itself (allocate and setMem default to rejecting, with javadoc explaining that wrap-only/read-only providers are the case they serve; destroy defaults to a no-op). SourceMemoryProvider is deleted, and CollectionDataMemoryProvider implements MemoryProvider directly, carrying its own getNumberSize as you suggested.

PR #373, opened 2026-07-31
Improvements to MemoryData IO

ashesfall:
Phase 18 of the setMem policy roadmap, covering items 1-3 of the ingest-layer plan (docs/plans/SETMEM_INGEST_LAYER.md).

MemoryData.read(ByteBuffer) is now the primary deserialization ingest form. The byte[] and InputStream overloads remain as thin adapters that wrap into a ByteBuffer, preserving the big-endian wire format exactly. GraphPersist reads query results through ByteBuffer.wrap rather than the byte[] form. The PackedCollection stream deserializer now delegates to read(InputStream) instead of a manual readDouble loop, since the wire formats are identical.

The four explicit-array MemoryData.setMem overloads — setMem(double[], int), setMem(double[], int, int), setMem(int, double[], int, int), and setMem(int, float[], int, int) — are removed. The varargs literal forms stay public and now terminate in two private setMemInternal methods that collapse the old delegation recursion into a single bounds check against the root reservation followed by the static Memory-level write. Callers migrated:

- PackedCollection.set/of route through the varargs form; clone now copies device-to-device via setFrom instead of a host round trip through toArray.
- Vector.setTo's cross-provider branch binds the varargs form.
- LocalExternalMemoryProvider and MemoryBankAdapter rewired internally.
- Llama2Weights.packComplex and the SAMEResampling/Moonbeam test data builders stage through NativeMemoryProvider ByteBuffer views and Bytes.of, so no host arrays flow through setMem.

Known residue: Java varargs binding still accepts an explicit array at a varargs call site. That path cannot be compile-blocked without losing literal support, so it remains enforced by SetMemLiteralsDetector (SETMEM_NON_LITERAL_ARGUMENT).

Ledger: three burn-down exclusions retired (Llama2Weights, OnnxFeatures, SAMEResamplingParityTest) leaving 18; baseline regenerated to 457 rows / 612 occurrences, a pure-removal diff. New tests: MemoryDataSerializationTest (round-trip and adapter equivalence, direct-buffer source) and MemoryDataViewWriteTest (root/view/nested-view writes, past-root rejection).

Verification: full clean reactor install plus test-compile sweep with -fae is green (an earlier incremental sweep missed warm-target modules, surfacing as a runtime NoSuchMethodError; the clean sweep closed that hole). Targeted runs all pass: engine/utils PackedCollectionTests, VectorMathTest, and the two new suites (20/20); studio/compose MoonbeamMidiTest (13/13). SetMemLiteralsDetector reports zero unbaselined violations and zero stale baseline rows; build validator passes all checks. The GraphPersist change is compile-verified by CI only, as flowtree/graphpersist does not build locally.


base/hardware/src/main/java/org/almostrealism/hardware/Hardware.java
```diff
@@ -632,6 +634,7 @@ private Hardware(String name, List<ComputeRequirement> reqs, Location location,
 		this.contextListeners = Collections.synchronizedList(new ArrayList<>());
 		this.contexts = new ArrayList<>();
 		this.nativeDirectBuffers = SystemUtils.isEnabled("AR_HARDWARE_NATIVE_DIRECT_BUFFERS").orElse(true);
+		this.nativeSharedMemory = nioMemory;
```
ashesfall: It no longer makes sense to name the parameter nioMemory, when NIO memory is available anytime and this flag now effectively refers only to the use of shared memory.

ashesfall: Renamed: the constructor parameter is now `sharedMemory` (both constructors, javadoc updated) and the static initializer's local is `sharedMem`, since the flag only governs whether cross-backend shared memory is enabled — NIO memory itself is available anytime through the lazily created staging provider. The old name was also shadowing the `nioMemory` provider field inside the constructor, which this clears up. Staged locally alongside the ByteBufferTransfer work; NativeBufferStagingModeTest and NativeBufferIngestTest pass against the change.

PR #377, opened 2026-08-01
Prevent direct MemoryData assignment from Java (Phase 20)

ashesfall:


engine/ml/src/test/java/org/almostrealism/ml/audio/SoftNormBottleneckTest.java
```diff
@@ -71,7 +71,7 @@ public void affineWithRunningStd() {
 
 		PackedCollection scale = new PackedCollection(shape(dim)).randnFill();
 		PackedCollection bias = new PackedCollection(shape(dim)).randnFill();
-		PackedCollection runningStd = new PackedCollection(shape(1)).fill(pos -> 2.5);
+		PackedCollection runningStd = new PackedCollection(shape(1)).fill(2.5);
```
ashesfall, line 74: pack(2.5) is simpler here

PR #381, opened 2026-08-02
Prevent direct MemoryData assignment from Java (Phase 21)

ashesfall:


engine/render/src/main/java/org/almostrealism/raytrace/AbsorberHashSet.java
```diff
@@ -1141,7 +1141,7 @@ public ShadableIntersection intersectAt(Producer<?> r) {
 
 			if (dist < Double.MAX_VALUE - 2 && dist > 0 && AbsorberHashSet.this.closest != null) {
 				di = new PackedCollection(1);
-				di.setMem(0, dist);
+				di.fill(dist);
```
ashesfall: why not di = pack(dist)?

engine/utils/src/main/java/org/almostrealism/util/TestFeatures.java
```diff
@@ -284,13 +286,50 @@ default void assertEquals(String msg, PackedCollection expected, PackedCollectio
...
 	default double compare(PackedCollection expected, PackedCollection actual) {
-		double exp[] = expected.toArray();
+		return compare(expected.toArray(), actual);
+	}
+
+	/**
+	 * Computes the average absolute difference between host reference values and a
+	 * device-computed {@link PackedCollection}.
+	 *
+	 * <p>Reference values for an assertion exist to be derived independently of the
```
ashesfall: This is just not true, we often do comparisons over values computed from different devices; or other variations. The compare method which takes PackedCollection for the expected values should not be removed, and in fact should actually be more widely used than it is.

engine/utils/src/main/java/org/almostrealism/util/TestFeatures.java
```diff
@@ -284,13 +286,50 @@ default void assertEquals(String msg, PackedCollection expected, PackedCollectio
...
+		return IntStream.range(0, expected.length)
+				.mapToDouble(i -> Math.abs(expected[i] - act[i]))
 				.average().orElseThrow();
 	}
 
+	/**
+	 * Materializes a function over every position of the given shape into a host array,
+	 * for use as reference values in an assertion.
+	 *
+	 * <p>This is deliberately host-side and deliberately not a {@link PackedCollection}.
```
ashesfall: If this is only ever used in combination with the compare method, then it would be better for the compare method to just have the option to accept a shape and function for the expectation computation. The existence of a method like this, that constructs a double[], creates a real risk of misbehavior from future agents that see it as a license to store tensors on the host for far more broad purposes than assertions/comarison.

engine/utils/src/test/java/org/almostrealism/collect/computations/test/TraversableDeltaComputationTests.java
```diff
@@ -655,7 +659,11 @@ public void divideProduct2() throws IOException {
 	@Test(timeout = 60000)
 	public void divideProduct3() throws IOException {
 		int c = 2;
-		divideProduct("divideProduct3", c, () -> new PackedCollection(c).fill(() -> Math.random() / 10.0));
+		divideProduct("divideProduct3", c, () -> {
+			PackedCollection o = new PackedCollection(c);
+			rand(o.getShape()).divide(10.0).into(o.traverseEach()).evaluate();
```
ashesfall: You don't need to create your own output and use into() this way, you can just return rand(o.getShape()).divide(10.0).evaluate()

ashesfall: Fixed, and swept the rest of the PR for it — there were 21 sites where I allocated a destination only to evaluate into it and then use it. They are now just `PackedCollection o = rand(shape(c)).divide(10.0).evaluate();` and equivalents.

A few needed a `reshape` after `evaluate()` because the producer is a flat index expression while the destination carried a multi-dimensional shape (e.g. `integers(...).evaluate().reshape(shape(n, hd))`) — that is the case where the destination was contributing shape information, so the shape is now stated explicitly instead of being implied by the allocation.

Left alone: seven pre-existing sites in `CollectionComputationTests` and `TraversableDeltaComputationTests` that pass a destination with a deliberate traversal (`.traverse(1)`, `.traverse(2)`, `new PackedCollection(shape(10), 1)`), since there the destination really is influencing the evaluation.

engine/utils/src/main/java/org/almostrealism/util/TestFeatures.java
ashesfall: You're right, and the javadoc I wrote was wrong — comparing two collections computed on different devices is exactly the case that makes the `PackedCollection` form valuable, and I described it as if the transfer were always pointless.

`compare(PackedCollection, PackedCollection)` is restored to its own implementation (it no longer delegates through a host-array overload) and the javadoc no longer disparages it. The `double[]` overload is gone entirely — see the reply on `reference` below.

engine/utils/src/main/java/org/almostrealism/util/TestFeatures.java
ashesfall: Done — `reference` and the `double[]` overload of `compare` are both gone, replaced by a single method that takes the shape and the function:

```java
default double compare(TraversalPolicy shape, Function<int[], Double> expected,
                       PackedCollection actual)
```

The expected values are evaluated inside it and never escape, so there is no host-array-of-tensor-values API for anything to reach for. The javadoc also points at `compare(PackedCollection, PackedCollection)` for the case where the expectation is itself a collection.

Your reasoning about future agents is well founded — I had put `reference` on `TestFeatures` rather than `TraversalPolicy` for exactly that concern, and then left the hazard in place anyway by exposing the array. Folding it into `compare` removes it properly.

`RotationTests` reads a bit better for it too, since the lambda now sits directly in the comparison rather than in a separate binding.

engine/render/src/main/java/org/almostrealism/raytrace/AbsorberHashSet.java
ashesfall: No reason — `di = pack(dist);` is better and it's now that. I was mechanically converting `setMem(0, x)` to `fill(x)` and kept the allocation that was already there instead of noticing the whole thing is one call.

PR #391, opened 2026-08-06
Prevent direct MemoryData assignment from Java (Phase 24)

ashesfall:


engine/audio/src/main/java/org/almostrealism/audio/BatchedPatternRenderer.java
```diff
@@ -120,6 +120,118 @@ public BatchedPatternRenderer(int n, int sourceLength, int targetLength,
...
+
+	/**
+	 * Returns one {@code [n]} row of {@link #sssScalars} as a bindable buffer. The row is a
+	 * view, so the kernel sees the same offset and size it would for an independent
+	 * allocation while the caller can write every column in one pass.
+	 *
+	 * @param column the column index within the scalar store
+	 * @return the bound buffer for that column
+	 */
+	private PackedCollection scalarColumn(int column) {
```
ashesfall: This should not have its own method, you keep falling for this and we need to introduce some kind of hook to detect it. I will state AGAIN:

If your collection is the right shape (*which it SHOULD be*) then this method is just restating collection.get(n). Even if it isn't the right shape, using PackedCollection::reshape before calling get() is far superior to using range() explicitly. It is called a collection for a reason: STOP recreating collection membership as a dedicated method all over the place. This is completely insane.

engine/audio/src/main/java/org/almostrealism/audio/BatchedPatternRenderer.java
```diff
@@ -120,6 +120,118 @@ public BatchedPatternRenderer(int n, int sourceLength, int targetLength,
...
+	}
+
+	/**
+	 * Returns one {@code [n]} row of {@link #percScalars} as a bindable buffer, on the same
+	 * basis as {@link #scalarColumn(int)}.
+	 *
+	 * @param column the column index within the percussion scalar store
+	 * @return the bound buffer for that column
+	 */
+	private PackedCollection percScalarColumn(int column) {
```
ashesfall: This should not have its own method, you keep falling for this and we need to introduce some kind of hook to detect it. I will state AGAIN:

If your collection is the right shape (*which it SHOULD be*) then this method is just restating collection.get(n). Even if it isn't the right shape, using PackedCollection::reshape before calling get() is far superior to using range() explicitly. It is called a collection for a reason: STOP recreating collection membership as a dedicated method all over the place. This is completely insane.

engine/audio/src/test/java/org/almostrealism/audio/benchmark/PatternRenderingFloorBenchmarkAdditional.java
```diff
@@ -69,50 +69,71 @@ public class PatternRenderingFloorBenchmarkAdditional extends PatternRenderingFl
...
-		final double[] tickStartOffset;
 
 		/**
-		 * Allocates all per-note arrays for the given total note count.
+		 * The per-note scalar set in the column-major {@code [COLUMNS, totalNotes]} layout
+		 * the batched dispatch binds, staged for a single upload per tick. Production
+		 * assembles exactly this shape — every bound scalar buffer is a row of one store —
+		 * so one staged block per tick is what the gather floor has to model.
+		 */
+		final double[] scalars;
```
ashesfall: The way this is used in the set method below makes me strongly suspect it should be a PackedCollection rather than a double[]. The comment here certainly does not defend the use of double[].

engine/audio/src/test/java/org/almostrealism/audio/benchmark/PatternRenderingFloorBenchmarkAdditional.java
```diff
@@ -676,17 +704,7 @@ private void runB1Gather(PackedCollection audioBuf, PackedCollection audioData,
...
-		PackedCollection.of(md.volDecay);
-		PackedCollection.of(md.volSustain);
-		PackedCollection.of(md.volRelease);
-		PackedCollection.of(md.filterAttack);
-		PackedCollection.of(md.filterDecay);
-		PackedCollection.of(md.filterSustain);
-		PackedCollection.of(md.filterRelease);
-		PackedCollection.of(md.automationLevel);
-		PackedCollection.of(md.tickStartOffset);
+		md.scalarStore.fill(md.scalars);
```
ashesfall: This could be setFrom if md.scalars was a PackedCollection, which it pretty clearly should be.

engine/audio/src/test/java/org/almostrealism/audio/BatchedSssChainTest.java
```diff
@@ -126,7 +124,7 @@ private Workload buildWorkload(BatchedPatternRenderer renderer) {
 			CollectionProducer merged = null;
 			for (int l = 0; l < LAYERS; l++) {
 				PackedCollection resampled =
-						renderer.buildResampleProducer(sourceByLayerNote[l][n], ratioValues[l][n])
+						renderer.buildResampleProducer(sourceByLayerNote[l][n], w.ratios[l].toDouble(n))
```
ashesfall: Are we unwrapping a PackedCollection to double just to wrap the value again inside the buildResampleProducer method?

engine/audio/src/test/java/org/almostrealism/audio/BatchedSssChainTest.java
```diff
@@ -141,26 +139,22 @@ private Workload buildWorkload(BatchedPatternRenderer renderer) {
...
-		double sumSqDiff = 0.0;
-		double sumSqRef = 0.0;
-		for (int i = 0; i < expected.length; i++) {
-			double diff = expected[i] - actual.toDouble(i);
-			sumSqDiff += diff * diff;
-			sumSqRef += expected[i] * expected[i];
-		}
-		double rms = Math.sqrt(sumSqDiff / expected.length);
-		double refRms = Math.sqrt(sumSqRef / expected.length);
+	private void assertRmsEquivalent(String label, PackedCollection expected, PackedCollection actual) {
```
ashesfall: If this is a standard audio testing feature, it belongs in the features interface for audio testing.

engine/audio/src/main/java/org/almostrealism/audio/BatchedPatternRenderer.java
ashesfall: Removed. `sssScalars` is already `[columns, n]`, so both accessors were `get(column)` written as arithmetic — `scalarColumn` and `percScalarColumn` are gone and the call sites index the store directly.

Swept the rest of the branch for the same shape and found five more, all now `get`: a per-note source batch, three coefficient-bank reads, a channel coefficient row in the adapter, and a feedback-matrix row. Two things that looked similar were left alone because the offset is not a member boundary — a prefix of a member, and a note's placement offset in an output window.

One correction worth recording: `get(i)` alone was wrong. It offsets along the traversal axis, and a `[rows, columns]` collection traverses axis 0 by default, where the single member is the whole collection — so the first version threw `Delegate offset is out of bounds` in two tests and silently returned the wrong view in a third. The idiom is `traverse(1).get(i)`, which `BatchedSssPlaybackTest` already had a comment about and I had read without registering.

On detecting it: added `.claude/hooks/warn-collection-membership.sh`, which flags `range(shape(X), i * X)` and any private `PackedCollection` accessor taking an `int` whose body is a `range()` call. It warns rather than blocks, since a range whose offset is not a multiple of the member size is legitimate. Backed by `.claude/hooks/rules/collection-membership.md`.

Also put it where it should have been found without a reviewer: `PackedCollection::range`'s javadoc now opens by saying it is not how to ask for a member, `get`'s documents the `traverse(1)` requirement and the exact error you get without it, and `compute/algebra/README.md` gained a "Members versus ranges" section — its own zero-copy example was doing this (`range(shape(100, 100), 10000)`), so that is fixed too. `QUICK_REFERENCE.md` and `packed-collection-examples.md` now show `traverse(1).get(i)` for members and describe `range` as a sub-region that is not one.

engine/audio/src/main/java/org/almostrealism/audio/BatchedPatternRenderer.java
ashesfall: Same fix — `percScalarColumn` is gone along with `scalarColumn`; see the reply on the other comment for the branch-wide sweep, the `traverse(1)` correction, and the hook and documentation changes.

engine/audio/src/test/java/org/almostrealism/audio/benchmark/PatternRenderingFloorBenchmarkAdditional.java
ashesfall: You are right, and the comment was answering a question you had not asked — it defended keeping the copy inside the timed region, which is about where the work happens, not about the type holding it.

`scalars` is a `PackedCollection` now and `set(column, note, value)` is gone. Every column was a uniform draw over its own interval, so the block is built on the device as one draw scaled and shifted per column, with the tick-start column floored afterwards because it is a sample index. The `column * totalNotes + note` arithmetic that `set` performed was also the membership pattern from the other thread, which is a second reason it should not have been there.

Two things fell out that are worth recording. The bounds are held as literal collections in column order rather than assigned per column by name — `setMem` takes literals only, so a `setBounds(spans, floors, COLUMN, floor, span)` helper is rejected by the policy detector, which I only found by running it. And with `set` gone the per-column name constants had no consumer at all, so ErrorProne failed on them; rather than keep them alive artificially I removed them and documented the column order on the field, keeping the two that are still read.

The benchmark runs and the whole file now holds no `double[]` except the per-density wall-clock timings.

engine/audio/src/test/java/org/almostrealism/audio/benchmark/PatternRenderingFloorBenchmarkAdditional.java
ashesfall: Done — `md.scalars` is a `PackedCollection` and this is `md.scalarStore.setFrom(0, md.scalars)`.

One consequence I do not want to leave implicit: the block is now device-resident, so this measures the copy rather than the host-side assembly production still performs ahead of it. The step stays inside the timed region for the reason the javadoc already gave — production's scalar set changes with the notes overlapping each window, so it cannot be staged ahead — but what the step costs is a different quantity than before. I have said so on the method rather than let the number be compared against earlier runs as though it were the same measurement.

engine/audio/src/test/java/org/almostrealism/audio/BatchedSssChainTest.java
ashesfall: Yes — `buildResampleProducer` took a `double` and immediately did `c(ratio)`, so the value was read out of the collection only to be wrapped straight back up.

There is a `Producer` overload now and the `double` form delegates to it, so the existing callers that genuinely hold a scalar are unaffected. Both per-note reference loops pass the ratio as it is held: `cp(w.ratios[l].get(n, shape(1)))`, which also uses the member accessor rather than an offset.

engine/audio/src/test/java/org/almostrealism/audio/BatchedSssChainTest.java
ashesfall: Moved to `FirFilterTestFeatures` as `assertRmsEquivalent`, beside the energy measurements it is built from.

It was already a second implementation: `BatchedSssPlaybackTest` had the same difference-RMS-against-reference-RMS comparison written out separately, and that is now the shared call too. `BatchedSssChainTest` picks it up by implementing the mixin.

It is expressed in terms of `differenceEnergy` and `energy` rather than repeating the sums, and it keeps the reference level and the ratio in the log, since a difference only means something against the signal it is a difference from.

